PR #77, opened 2017-04-16
fix default exports, use jsx components

laggingreflex:
This does two things:

* Simpler syntax: `export { default as Button } from './Button/Button.jsx';`

* Uses `Button/Button.jsx` instead of `Button/index.js` (which is created by `compileComponents.js`). It lets you run `webpack --watch` without `compileComponents`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,36 +1,17 @@
-import Button from './Button';
-import Checkbox from './Checkbox';
-import Fab from './Fab';
-import Icon from './Icon';
-import Switch from './Switch';
-import Snackbar from './Snackbar';
-import Radio from './Radio';
-import Textfield from './Textfield';
-import Select from './Select';
-import Card from './Card';
-import Elevation from './Elevation';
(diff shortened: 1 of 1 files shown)


prateekbh:
I totally agree about adding a simpler syntax to main index.js.
But there is a reason why I import that from index.js and not particular Component.jsx.
Imagine you are using building a project, which works absolutely fine, then you add `preact-material-components`. 

Suddenly you realize that to use these components, you'll need to change your webpack config or gulp config or something as they use some more stuff from ES6/7/.... This might become frustrating for anyone using these, hence I compile Component.js to a basic ES6 file and put them in index.js.

For people using ES5 a UMD version is kept in dist under `main` key.
And using `module` key a basic ESM version is pointed to all index.js

prateekbh:
I also totally support that compileComponents should become part of webpack process itself, May be we can just create a small webpack plugin out  of it just like I did for `CssMigrationWebpackPlugin`

PR #96, opened 2017-04-25
Improvements to Textfield component

IrregularShed:
* Removes the hardcoded `type="text"`
* Adds `type` as a prop (for passwords, dates, etc)
* Removes duplicate `className`

### DIFF
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -5,6 +5,7 @@ import { MDCTextfield } from "@material/textfield/";
 /**
  * @prop fullwidth = false
  * @prop multiline = false
+ * @prop type = 'text'
  */
 export default class Textfield extends MaterialComponent {
   constructor() {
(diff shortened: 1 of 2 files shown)


Textfield/Textfield.jsx
```diff
@@ -13,7 +14,7 @@ export default class Textfield extends MaterialComponent {
     this.state = {
       showFloatingLabel: false
     };
-    this._mdcProps = ["fullwidth", "multiline"];
+    this._mdcProps = ["fullwidth", "multiline", "type"];
```
prateekbh: Adding a prop in this index, puts it in a loop to be constructed as a class.
https://github.com/prateekbh/preact-material-components/blob/master/MaterialComponent.js#L35-L37

You can skip this and still it will as desired.

P.S.: only Boolean kind of props land here which end up becoming classes for components root elements.

IrregularShed: I should've remembered that from before. Never mind - I'll remove it now.

PR #99, opened 2017-04-26
Documentation for Textfield change

IrregularShed:
* Adds doc and demo for using `props.type`

The styling of `TextfieldPage` seems to have gone a bit awry after `lint-staged` crept in. If you prefer I'll redo, although I think the `package.json` scripts need checking - the `lint` script is clearly wrong as `eslint` isn't a dependency and `src/` doesn't exist!


prateekbh:
@IrregularShed can you update in package.json to version 1.0.1?

IrregularShed:
Done - I've updated the `start` and `build` scripts whilst there

prateekbh:
@IrregularShed done thanks 👍

Issue #129, opened 2017-05-23
HMR and event handlers


cedric-marcone:
Hey ! ...me again :)

In https://github.com/developit/preact-cli HMR is active and it doesn't play well with the event handlers you're binding.

I think the reason is you're attaching them in ```componentDidMount``` but they are never cleared in ```componentWillUnmount```.

Do you accept PRs ?

prateekbh:
yes I love PRs... please feel free to go ahead and open one

cedric-marcone:
Hmm, I'm banging my head on this one !! :)
Even when I "unlisten" the event handler in componentWillUnmount it continues to be called once per reload via HMR...
I'll continue to investigate tonight and if I fail miserably I'll commit my changes in the fork tomorrow morning.

If you have the opportunity to review it and give me some insight, I'll be grateful :)  

I keep you updated.

cedric-marcone:
OK, I really don't know what's going on here, I'm way too noob on preact / HMR...
Could you take a look at https://github.com/cedric-marcone/preact-material-components/commit/96a2f0dc59907f4edc1df94d761b912aa89f07d7 and tell me what you think ?
Nothing big going on there, but it really doesn't work : every time I update the code of a  component displaying a ```Select``` an onChange handler is added, even if it seems to be correctly unregistered.

prateekbh:
Ha no probs, i'll take in the morning

prateekbh:
@cedric-marcone can u do me a favor... Can u set up hmr in a separate repo for me.
And add the instructions to run hmr in readme?

cedric-marcone:
Here it is :
https://github.com/cedric-marcone/preact-material-components-test

To setup HMR I didn't do anything more than using preact-cli, the interesting piece is in ```routes/home/index.js```

The README explains how to reproduce the bug.

prateekbh:
thanks will look into it

prateekbh:
ok cool had a look, so its not about `unlisten` as the root element itself gets destroyed. But the MDCComponent itself gets initialized multiple times and thus emitting multiple events.
Just in case if you wanna PR this, on the eaxct same place you called `unlisten` you need to replace the line to 
```js
this.MDComponent.destroy();
``` 
and this should fix it for you

cedric-marcone:
Of course !! That makes sense.
I prepare a PR.
Thanks a lot.

cedric-marcone:
Thanks for the merge !

[...]

PR #132, opened 2017-05-24
Cleanup MDComponents allocations and event handlers

cedric-marcone:
As seen in #129, here is the PR.
Just destroying the underlying ```MDComponent``` in ```componentWillUnmount``` was not enough. I had to remove event handlers as well in ```Select```and ```Drawer``` components.
Thanks.

### DIFF
--- a/Checkbox/Checkbox.jsx
+++ b/Checkbox/Checkbox.jsx
@@ -14,7 +14,7 @@ export default class Checkbox extends MaterialComponent {
     this.MDComponent = new MDCCheckbox(this.control);
   }
   componentWillUnmount() {
-    this.MDComponent.destroy();
+    this.MDComponent.destroy && this.MDComponent.destroy();
   }
   materialDom(allprops) {
     const { className, ...props } = allprops;
--- a/Checkbox/Checkbox.jsx
+++ b/Checkbox/Checkbox.jsx
@@ -13,6 +13,9 @@ export default class Checkbox extends MaterialComponent {
   componentDidMount() {
     this.MDComponent = new MDCCheckbox(this.control);
   }
+  componentWillUnmount() {
+    this.MDComponent.destroy();
+  }
   materialDom(allprops) {
     const { className, ...props } = allprops;
     return (
--- a/Drawer/Drawer.jsx
+++ b/Drawer/Drawer.jsx
@@ -29,6 +29,7 @@ class TemporaryDrawer extends MaterialComponent {
   componentWillUnmount() {
     this.MDComponent.unlisten("MDCTemporaryDrawer:close", this._close);
     this.MDComponent.unlisten("MDCTemporaryDrawer:open", this._open);
+    this.MDComponent.destroy && this.MDComponent.destroy();
   }
   materialDom(props) {
     return (
@@ -136,6 +137,7 @@ class PersistentDrawer extends MaterialComponent {
   componentWillUnmount() {
     this.MDComponent.unlisten("MDCPersistentDrawer:close", this._close);
     this.MDComponent.unlisten("MDCPersistentDrawer:open", this._open);
+    this.MDComponent.destroy && this.MDComponent.destroy();
   }
   materialDom(props) {
     return (
--- a/Drawer/Drawer.jsx
+++ b/Drawer/Drawer.jsx
@@ -8,20 +8,27 @@ class TemporaryDrawer extends MaterialComponent {
   constructor() {
     super();
     this.componentName = "temporary-drawer";
+    this._open = this._open.bind(this);
+    this._close = this._close.bind(this);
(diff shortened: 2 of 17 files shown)


Checkbox/Checkbox.jsx
```diff
@@ -13,6 +13,9 @@ export default class Checkbox extends MaterialComponent {
   componentDidMount() {
     this.MDComponent = new MDCCheckbox(this.control);
   }
+  componentWillUnmount() {
+    this.MDComponent.destroy();
```
prateekbh: since destroy is something that is not publicly documented, do you think it is better to safe guard it with an `&&` condition?

```js
```

Drawer/Drawer.jsx
```diff
@@ -8,20 +8,27 @@ class TemporaryDrawer extends MaterialComponent {
...
-      if (this.props.onClose) {
-        this.props.onClose();
-      }
-    });
+    this.MDComponent.listen("MDCTemporaryDrawer:open", this._open);
+    this.MDComponent.listen("MDCTemporaryDrawer:close", this._close);
+  }
+  componentWillUnmount() {
+    this.MDComponent.unlisten("MDCTemporaryDrawer:close", this._close);
+    this.MDComponent.unlisten("MDCTemporaryDrawer:open", this._open);
```
prateekbh: shouldn't we do a destroy here as well?

Drawer/Drawer.jsx
```diff
@@ -108,20 +115,27 @@ class PersistentDrawer extends MaterialComponent {
...
-      if (this.props.onClose) {
-        this.props.onClose();
-      }
-    });
+    this.MDComponent.listen("MDCPersistentDrawer:open", this._open);
+    this.MDComponent.listen("MDCPersistentDrawer:close", this._close);
+  }
+  componentWillUnmount() {
+    this.MDComponent.unlisten("MDCPersistentDrawer:close", this._close);
+    this.MDComponent.unlisten("MDCPersistentDrawer:open", this._open);
```
prateekbh: same `destroy` missing comment

Select/Select.jsx
```diff
@@ -7,18 +7,24 @@ class Select extends MaterialComponent {
...
   }
   componentDidMount() {
     if (!this.props.basic) {
       this.MDComponent = new MDCSelect(this.control);
-      this.MDComponent.listen("MDCSelect:change", () => {
-        if (this.props.onChange) {
-          this.props.onChange();
-        }
-      });
+      this.MDComponent.listen("MDCSelect:change", this._changed);
```
prateekbh: cant we simply do a 
```js
this.MDComponent.listen("MDCSelect:change", this._changed.bind(this));
```

this will not add any extra steps anywhere

Checkbox/Checkbox.jsx
cedric-marcone: Ah ! I was not aware ```destroy``` wasn't part of the public API : you're absolutely right.
I'll safe guard every call to this method.

Drawer/Drawer.jsx
cedric-marcone: Yes ! Definitely.

Drawer/Drawer.jsx
cedric-marcone: ok

cedric-marcone:
On another topic, any idea why the CI is failing on missing components files ?
I can't repro. locally and you didn't seem to have the problem with your last commit.

prateekbh:
@cedric-marcone i'll do a pr tomorrow, please do a pull and it'll streamline all the commit/ci process  for u

prateekbh:
@cedric-marcone you can commit and push your changes after taking a pull, don't worry about index.js files. Prettier will take care of them.

Also once the next version of prettier is released i'll add it to webpack script so that no one sees the extra list of files

P.S. please run `webpack` before sumitting

Drawer/Drawer.jsx
prateekbh: the `destroy` call is missing here

prateekbh:
btw, your CI passed with the new changes 😃

cedric-marcone:
Ah ! I didn't see ```attachTo``` was really instanciating a Component.
It seems they're several calls to it :
```
Drawer/Drawer.jsx:    this.MDComponent = MDCTemporaryDrawer.attachTo(this.control);
Drawer/Drawer.jsx:    this.MDComponent = MDCPersistentDrawer.attachTo(this.control);
Radio/Radio.jsx:    this.MDComponent = MDCRadio.attachTo(this.control);
Snackbar/Snackbar.jsx:    this.MDComponent = MDCSnackbar.attachTo(this.control);
```
I'll tackle it, right now.

PR #133, opened 2017-05-25
adding custom details to event

### LINKED ISSUE #128: Select onChange handler
Hello again, 

Would you mind adding selectedOptions and selectedIndex when calling the event handler ?

```js
this.props.onChange(
  this.MDComponent.selectedOptions,
  this.MDComponent.selectedIndex
);
````

It could let us write more generic code, without ```refs```tricks ;)

```js
selectedChanged = (name) => (values, i) => {
   console.log(name, i, values)
   this.setState({ [name]: i })
}

render() {
 return (
    <Select onChange={this.selectedChanged('chosenOption')}>
    ...
  )
}
```

prateekbh:
this adds the selected index and selected options to the event.'
Fixes #128


prateekbh:
@cedric-marcone PTAL, if this satisfies what u were looking for

sample/home.jsx
```diff
@@ -151,9 +151,9 @@ export default class Home extends Component {
             this.presel = presel;
```
cedric-marcone: Shouldn't we remove the "ref trick" in the example now that it's not needed anymore to extract values ?

cedric-marcone:
LGTM for the update : perfect and future proof, thanks.
But, I added a comment regarding the documentation example for the Select component.

sample/home.jsx
prateekbh: i'll update the gh-pages

PR #143, opened 2017-06-02
Fix samples of various components

filoozom:
This fixes some imports in various components and adds styles where they are needed. For example, we need ```Button/style.css``` in ```CardsPage``` because ```Card.Action``` needs it.


filoozom:
Shouldn't there be a npm script running ```webpack -p```? Maybe edit the build script or create ```build:prod```, would be easier for newcomers.

prateekbh:
@filoozom definitely, will do it while upgrading it repo to mdc 0.12.1

PR #161, opened 2017-06-25
Add Slider component

### LINKED ISSUE #160: Add Slider component
Consider adding the mdc-slider https://github.com/material-components/material-components-web/tree/master/packages/mdc-slider

MobiusHorizons:
Fixes #160

### DIFF
--- a/Slider/Slider.jsx
+++ b/Slider/Slider.jsx
@@ -1,0 +1,79 @@
+import { h } from "preact";
+import MaterialComponent from "../MaterialComponent";
+import { MDCSlider } from "@material/slider";
+
+/**
+ * @prop disabled = false
+ */
+export default class Slider extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "slider";
+    this._mdcProps = [];
+    this._onChange = this._onChange.bind(this);
+    this._onInput = this._onInput.bind(this);
+  }
+  _onChange() {
+    if (this.props.onChange) {
+      this.props.onChange(this.MDComponent.value);
+    }
+  }
+  _onInput() {
+    if (this.props.onInput) {
+      this.props.onInput(this.MDComponent.value);
+    }
+  }
+  componentDidMount() {
+    this.MDComponent = new MDCSlider(this.base);
+    this.MDComponent.listen("MDCSlider:change", this._onChange);
+    this.MDComponent.listen("MDCSlider:input", this._onInput);
+  }
+  componentWillUnmount() {
+    this.MDComponent.unlisten("MDCSlider:change", this._onChange);
+    this.MDComponent.unlisten("MDCSlider:input", this._onInput);
+    this.MDComponent.destroy && this.MDComponent.destroy();
+  }
+  materialDom(allprops) {
+    const {
+      disabled = false,
+      min = 0,
+      max = 100,
+      value,
+      step,
+      tabindex = 0,
+      ...props
+    } = allprops;
+
+    let dynamicProps = {};
+    if (disabled) {
+      dynamicProps["aria-disabled"] = true;
+    }
+
+    if (step) {
+      dynamicProps["data-step"] = step;
+    }
+
+    return (
+      <div
+        tabindex={tabindex}
+        role="slider"
+        aria-valuemin={min}
+        aria-valuemax={max}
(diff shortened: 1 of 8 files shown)


Slider/index.js
```diff
@@ -0,0 +1,118 @@
...
+import { MDCSlider } from "@material/slider";
+
+/**
+ * @prop disabled = false
+ */
+export default class Slider extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "slider";
+    this._mdcProps = ["disabled", "min", "max", "value", "step"];
```
prateekbh: So the way _mdcProps works is:
if user gives the value as true/false then these will be converted to specific `mdc_componentname--classname`

https://github.com/prateekbh/preact-material-components/blob/master/MaterialComponent.js#L35-L37

So I dont this these are needed here

Slider/Slider.jsx
```diff
@@ -0,0 +1,79 @@
...
+import { MDCSlider } from "@material/slider";
+
+/**
+ * @prop disabled = false
+ */
+export default class Slider extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "slider";
+    this._mdcProps = ["disabled", "min", "max", "value", "step"];
```
prateekbh: So the way _mdcProps works is:
if user gives the value as true/false then these will be converted to specific mdc_componentname--classname

https://github.com/prateekbh/preact-material-components/blob/master/MaterialComponent.js#L35-L37

So I dont this these are needed here

Slider/Slider.jsx
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    if (step) {
+      dynamicProps["data-step"] = step;
+    }
+
+    return (
+      <div
+        class={className + " mdc-slider"}
+        tabindex="0"
```
prateekbh: can we derive 0 as override-able prop?

Slider/Slider.jsx
MobiusHorizons: oh ok, i'll remove them.

Slider/Slider.jsx
```diff
@@ -0,0 +1,79 @@
...
+      dynamicProps["aria-disabled"] = true;
+    }
+
+    if (step) {
+      dynamicProps["data-step"] = step;
+    }
+
+    return (
+      <div
+        class={className + " mdc-slider"}
```
MobiusHorizons: got it.

Slider/Slider.jsx
MobiusHorizons: I'll add it in.

Slider/Slider.jsx
```diff
@@ -0,0 +1,80 @@
...
+      dynamicProps["aria-disabled"] = true;
+    }
+
+    if (step) {
+      dynamicProps["data-step"] = step;
+    }
+
+    return (
+      <div
+        class={className + ""}
```
prateekbh: I dont think a separate className is required at all ... dynamicProps will put it here

Slider/Slider.jsx
```diff
@@ -0,0 +1,80 @@
...
+        tabindex={tabindex}
+        role="slider"
+        aria-valuemin={min}
+        aria-valuemax={max}
+        aria-valuenow={value}
+        aria-label="Select Value"
+        {...dynamicProps}
+      >
```
prateekbh: also add `...props` here, so that anything extra a User wanna add, they can

MobiusHorizons: ok, i was sort of going back and forth on that one. If you are cool with that, i'll throw it in

Slider/Slider.jsx
MobiusHorizons: oh, i misunderstood this one, let me clean that up.

MobiusHorizons:
ok, this should be good now. I also rebased and squashed so it's just one commit in the history.

prateekbh:
hey, I am publishing this now, do you want to add this to `gh-pages` branch or shall I go ahead and do it?

MobiusHorizons:
I can probably do it this weekend, but if you get to it first, I certainly wouldn't mind. 
Thank you for the great library!!

prateekbh:
i've update the docs as well :)

PR #172, opened 2017-07-12
remove autogenerated files from git

reznord:
add index.js (autogenerated files) to .gitignore


prateekbh:
this alone wont be enough as the idex.js files inside the folders are already committed, u'll need to delete them along with this change.

PR #174, opened 2017-07-12
Icon toggle

### LINKED ISSUE #148: Implement icon-toggle
component `icon-toggle` is missing from this repo and needs to be implemented in the same manner as the rest of the components

reznord:
Icon-toggle component added. Fixes #148

### DIFF
--- a/sample/home.jsx
+++ b/sample/home.jsx
@@ -90,8 +90,8 @@ export default class Home extends Component {
             tabindex="0"
             aria-pressed="false"
             aria-label="Add to favorites"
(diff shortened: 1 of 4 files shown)


sample/home.jsx
```diff
@@ -83,6 +84,13 @@ export default class Home extends Component {
           <Switch />
         </div>
 
+        <div>
+          <Fab mini={true} ripple={true}>
+            <IconToggle>share</IconToggle>
```
prateekbh: can you put it outside fab? and pass props displaying the actual toggling?

PR #180, opened 2017-07-22
Fix LayoutGrid sample code

filoozom:
Adds the ```LayoutGrid.Inner``` component to the sample code.

(My bad, added a fix for the ```Card``` sample code in the same branch instead of create a new one)


prateekbh:
is `package-lock.js` really needed? I am unaware of the usage of this

filoozom:
It's not, just forgot to unstage the file created by ```npm``` since version 5 automatically. Deleted it!

prateekbh:
can you put a snapshot of whats the change?
I cant really figure out

filoozom:
Sorry, not quite sure what you're asking. You mean for the ```Card``` or the ```LayoutGrid``` sample code fix?

prateekbh:
what changes on the page appearance?

filoozom:
In the documentation of the ```LayoutGrid``` component, ```LayoutGrid.Inner``` was missing from the components list and wasn't used in the sample code. Without ```LayoutGrid.Inner```, the grid doesn't work (every cell is on its own line and doesn't respect the defined ```cols```).

I just added ```LayoutGrid.Inner``` to the components list and in the sample code as a parent from de ```LayoutGrid.Cell``` components.

PR #183, opened 2017-07-23
Added in onChange event and className extending

### LINKED ISSUE #179: IconToggle should have `onChange` handler wired up.
The IconToggle component currently can have it's state set, and will update on it's own. However for real use in a preact app, it will need to expose the `MDCIconToggle:change` event as an `onChange` event.

MobiusHorizons:
Fixes #179

### DIFF
--- a/IconToggle/IconToggle.jsx
+++ b/IconToggle/IconToggle.jsx
@@ -9,12 +9,20 @@ export default class IconToggle extends MaterialComponent {
   constructor() {
     super();
     this.componentName = "icon-toggle";
+    this._onChange = this._onChange.bind(this);
+  }
+  _onChange(e) {
+    if (this.props.onChange) {
+      this.props.onChange(e);
+    }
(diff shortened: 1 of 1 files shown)


IconToggle/IconToggle.jsx
```diff
@@ -9,12 +9,20 @@ export default class IconToggle extends MaterialComponent {
...
+      this.props.onChange(e);
+    }
   }
   componentDidMount() {
     this.MDComponent = new MDCIconToggle(this.control);
+    this.MDComponent.listen("MDCIconToggle:change", this._onChange);
   }
   componentWillUnmount() {
     this.MDComponent.destroy && this.MDComponent.destroy();
+    this.MDComponent.unlisten("MDCIconToggle:change", this._onChange);
```
prateekbh: not sure but do you think we should first unlisten and then destroy?

PR #189, opened 2017-07-23
Improve Textfield component

### LINKED ISSUE #188: Improve Textfield component
The ```Textfield``` component is missing a lot of features. For reference, [this is the GitHub documentation](https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield).

Missing features include:
- [ ] ```mdc-textfield--disabled``` class on disabled fields
- [ ] Dense
- [ ] Helper text (that can be persistent or not)
- [ ] The label is only floating when a value is set on the field, but it should float on focus

filoozom:
Maybe shouldn't have created an issue (fixes #188), but didn't know I'd have time to work on this straight away. This is a work in progress to add all the missing functionality to ```Textfield```.

- [x] Add ```mdc-textfield--disabled``` class on disabled fields
- [x] Add dense
- [x] The label is only floating when a value is set on the field, but it should float on focus
- [x] Add helper text (that can be persistent or not)
- [x] Add text field boxes
- [x] Avoid Flash Of Un-styled Content (FOUC)
- [x] CSS only textfield

I think we should also implement some way to recursively remove all ```_mdcProps``` from child elements. For example, in ```Textfield``` the ```<input>``` or ```<textarea>``` get all ```{...props}```, and the ```_mdcProps``` aren't removed from them because they aren't the parent element, if I understand it correctly.

### DIFF
--- a/MaterialComponent.js
+++ b/MaterialComponent.js
@@ -20,6 +20,12 @@ export default class MaterialComponent extends Component {
     // The final class name given to the dom
     this.classText = "";
   }
+  static uid() {
+    if (!this.uidCounter) {
+      this.uidCounter = 0;
+    }
+    return ++this.uidCounter;
+  }
   attachRipple() {
     if (this.props.ripple && this.control) {
       MDCRipple.attachTo(this.control);
--- a/MaterialComponent.js
+++ b/MaterialComponent.js
@@ -20,12 +20,6 @@ export default class MaterialComponent extends Component {
     // The final class name given to the dom
     this.classText = "";
   }
-  static uid() {
-    if (!this.uidCounter) {
-      this.uidCounter = 0;
-    }
-    return ++this.uidCounter;
-  }
   attachRipple() {
     if (this.props.ripple && this.control) {
       MDCRipple.attachTo(this.control);
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -167,7 +167,7 @@ class Textfield extends Component {
               {props.helptext}
             </Helptext>}
         </div>
-      : <TextfieldInput {...props} id={tfId} className={className} />;
+      : <TextfieldInput {...props} className={className} />;
   }
 }
 
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -79,13 +79,11 @@ class TextfieldInput extends MaterialComponent {
         {props.multiline
           ? <textarea
               className="mdc-textfield__input"
-              aria-controls={props.id + "-helptext"}
               {...props}
             />
           : <input
               type={props.type || "text"}
               className="mdc-textfield__input"
-              aria-controls={props.id + "-helptext"}
               {...props}
             />}
         {props.label &&
@@ -162,7 +160,7 @@ class Textfield extends Component {
             <label for={props.id}>
               {props.cssLabel || `${props.label}: `}
             </label>}
-          <TextfieldInput {...props} />
+          <TextfieldInput {...props} aria-controls={props.helptext && props.id + "-helptext"} />
           {props.helptext &&
             <Helptext id={props.id + "-helptext"} {...helptextProps}>
               {props.helptext}
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -84,7 +84,14 @@ class TextfieldInput extends MaterialComponent {
  */
 export default class Textfield {
   constructor() {
-    this.id = MaterialComponent.uid();
+    this.id = Textfield.uid();
+  }
+
+  static uid() {
+    if (!this.uidCounter) {
+      this.uidCounter = 0;
+    }
+    return ++this.uidCounter;
   }
 
   render(allprops) {
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -11,15 +11,9 @@ export default class Textfield extends MaterialComponent {
   constructor() {
     super();
     this.componentName = "textfield";
-    this.state = {
-      showFloatingLabel: false
-    };
-    this._mdcProps = ["fullwidth", "multiline"];
+    this._mdcProps = ["fullwidth", "multiline", "dense", "disabled"];
   }
   componentDidMount() {
-    this.setState({
-      showFloatingLabel: true
-    });
     this.MDComponent = new MDCTextfield(this.control);
   }
   componentWillUnmount() {
@@ -41,7 +35,6 @@ export default class Textfield extends MaterialComponent {
               {...props}
             />}
         {props.label &&
-          this.state.showFloatingLabel &&
           <label className={labelClass.join(" ")}>
             {props.label}
           </label>}
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -65,7 +65,11 @@ class TextfieldInput extends MaterialComponent {
     this.MDComponent.destroy && this.MDComponent.destroy();
   }
   materialDom(allprops) {
-    const { className, ...props } = allprops;
+    let { className, ...props } = allprops;
+
+    if (props.value && this.state.showFloatingLabel) {
+      className = [className, "mdc-textfield--upgraded"].join(" ");
+    }
 
     return (
       <div className={className} ref={control => (this.control = control)}>
@@ -83,7 +87,10 @@ class TextfieldInput extends MaterialComponent {
             />}
         {props.label &&
           this.state.showFloatingLabel &&
-          <Label for={props.id}>
+          <Label
+            for={props.id}
+            className={props.value && "mdc-textfield__label--float-above"}
+          >
             {props.label}
           </Label>}
       </div>
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -32,7 +32,7 @@ class TextfieldInput extends MaterialComponent {
   constructor() {
     super();
     this.componentName = "textfield";
-    this._mdcProps = ["fullwidth", "multiline", "dense", "disabled"];
+    this._mdcProps = ["fullwidth", "multiline", "dense", "disabled", "box"];
     this.state = {
       showFloatingLabel: false
     };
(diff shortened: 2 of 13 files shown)


filoozom:
So, I implemented a basic helper text, but there are two problems:

1. We need to set an ID to the component: ```<Textfield helptext="Text" id="id" />``` because we need it for the ```aria-controls``` on the ```input``` / ```textarea``` and on the ```<Helptext>```.
2. We have an useless outer ```div``` when ```helptext``` is not used.
3. The outer ```div``` has an ```mdc-``` class because of ```MaterialComponent```

Potential fixes:
1. I thaught about generating random IDs, but is that really a good idea?
2. I would have like for the ```Helptext``` to be an exported component that one could put independently from a ```Textfield``` because I feel like it would be a lot cleaner, both for props and for ```Textfield```'s ```materialDom``` function, but I don't know if it would work because I think that ```componentDidMount``` from the ```Textfield``` might fire before the ```Helptext``` component renders, which means that the ```Helptext``` would never be linked to the input as per https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/index.js#L36.
3. Maybe we should not add ```mdc-``` if ```componentName``` is empty.

Any ideas @prateekbh?

filoozom:
Ended up implementing something similar to https://github.com/developit/preact-mdl/blob/master/src/index.js#L15-L19 / https://github.com/developit/preact-mdl/blob/master/src/index.js#L854 for the automatic IDs.

prateekbh:
All of your recent work has been amazing!!! 💯 💯 💯 
I will merge this and publish a new version soon. Will you be awesome/kind enough to update `gh-pages` as well?

filoozom:
I will for sure 😃

MaterialComponent.js
```diff
@@ -20,6 +20,12 @@ export default class MaterialComponent extends Component {
     // The final class name given to the dom
     this.classText = "";
   }
+  static uid() {
```
prateekbh: can we move this to textfield instead, cuz no other component use this

filoozom: Sure. Thought I'd put it there for future-proofing because other elements might need it. But for now it's probably better to keep it in ```Textfield```.

Textfield/Textfield.jsx
```diff
@@ -3,49 +3,120 @@ import MaterialComponent from "../MaterialComponent";
...
     super();
     this.componentName = "textfield";
-    this.state = {
-      showFloatingLabel: false
-    };
-    this._mdcProps = ["fullwidth", "multiline"];
+    this._mdcProps = ["fullwidth", "multiline", "dense", "disabled"];
   }
   componentDidMount() {
-    this.setState({
```
prateekbh: this check was in place because floating labels(the way we are providing them) are only possible for JS enabled clients.

for a non jS client they should be outside the container

https://github.com/material-components/material-components-web/tree/master/packages/mdc-textfield#multi-line---gracefully-degraded

Textfield/Textfield.jsx
```diff
@@ -3,49 +3,120 @@ import MaterialComponent from "../MaterialComponent";
...
+    }
+    return ++this.uidCounter;
+  }
+
+  render(allprops) {
+    const {
+      id,
+      className,
+      helptextPersistent,
+      helptextValidationMsg,
```
prateekbh: would it be better to have full `helptextValidationMessage`?

filoozom: As you prefer, I chose ```helptextValidationMsg``` because of the name of the class, which is ```mdc-textfield-helptext-validation-msg```, so it was just to stay as close to original as possible, but I can change it if you want.

prateekbh: oh no, it makes sense now. lets keep it this way

Textfield/Textfield.jsx
prateekbh: please revert this change

filoozom: I'm not at home right now, but I'm pretty sure that it didn't work before either. The documentation is broken right now for example, isn't it? I'll have a look at it in a few hours.

prateekbh: sure no probs, so the idea is: `componentDidMount` executes only on client, hence when jS is present hence the check

filoozom: Right, I got that, but the issue is that the state in ```componentDidMount``` is changed after ```this.MDComponent``` is instanciated, which means that the label only appears after ```MDCTextfield``` has set the label to ```null``` because it didn't exist, which means that ```material-components-web``` doesn't update the label and add the floating class on focus.
I'm on the phone, not quite sure if what I'm writing makes sense, but I'm pretty sure that it didn't work before. Just check on the documentation, you'll see that the label never floats up.

filoozom:
I think this should work. Now we just call ```this.MDComponent = new MDCTextfield(this.control);``` once the rerender with the label is done, which is the important part!

filoozom:
I now have somewhat of a working example that works for both CSS-only and JS. Also added ```cssLabel``` to set a different label when it's CSS-only. I think this might be useful for other languages for example. I now wonder if I shouldn't just add a ```Textfield.CSSLabel``` instead though. Maybe even a ```no-js``` class or something like that.
Just need to tidy it up somehow for the ```showFloatingLabel``` state. Do you have any ideas?

Textfield/Textfield.jsx
```diff
@@ -1,50 +1,156 @@
...
+    const tfId = id || "tf-" + this.id;
+
+    // Helper text
+    const helptextProps = {
+      persistent: helptextPersistent,
+      "validation-msg": helptextValidationMsg
+    };
+
+    return (
+      <div className={className}>
```
prateekbh: this will result into div->div->input which breaks us away from mdc-web
put the `{showFloatingLabel &&` check outside and put the extra div only incase of no js

Textfield/Textfield.jsx
```diff
@@ -1,50 +1,156 @@
...
+        {props.label &&
+          !showFloatingLabel &&
+          <label for={tfId}>
+            {props.cssLabel || `${props.label}: `}
           </label>}
+        <TextfieldInput {...props} id={tfId} />
```
prateekbh: incase I dont want helper text or anything else, `className` should reach to the input tag

Textfield/Textfield.jsx
filoozom: But also when there is a help text then I guess.

prateekbh:
>One other somewhat "breaking" change is that Textfield is now always inside a <div> that gets the className instead of the TextfieldInput that did before. But this actually makes it more consistent with the labels and text helpers.

we should make the div has something like `className-container`.

lets make sure that end result markup should be same as mdc-web

filoozom:
Ok, so now I added ```Textfield.Helptext```, so that we can do something like:
```html
<Textfield label="Username" id="textfield" />
<Textfield.Helptext id="textfield-helptext">
	Test
</Textfield.Helptext>
```
but the ```id``` would have to be set accordingly. At least like that we don't have the outer ```<div>```. What do you prefer?

Textfield/Textfield.jsx
```diff
@@ -1,51 +1,176 @@
...
+            <label for={tfId}>
+              {props.cssLabel || `${props.label}: `}
+            </label>}
+          <TextfieldInput {...props} id={tfId} />
+          {props.helptext &&
+            <Helptext id={tfId + "-helptext"} {...helptextProps}>
+              {props.helptext}
+            </Helptext>}
+        </div>
+      : <TextfieldInput {...props} id={tfId} className={className} />;
```
prateekbh: lets not put auto generated Id for this case

filoozom: Right, that one didn't quite make sense, my bad.

filoozom:
Do you have a more elegant solution to replace the double
```
this.setState({
      showFloatingLabel: true
    });

File: Textfield/Textfield.jsx [hunk 7]
```diff
@@ -167,7 +167,7 @@ class Textfield extends Component {
               {props.helptext}
             </Helptext>}
         </div>
-      : <TextfieldInput {...props} id={tfId} className={className} />;
```
prateekbh: since you deconstructed `id` props wont have it. and now even if I give an id it wont be applied.

better would be to put it back in props I guess

filoozom: My bad, it's early. Will fix this as soon as I get my hands on a computer.

filoozom:
Well, I guess I should've put ```<TextfieldInput {...props} className={className} id={id} />``` for consistency. Or do you prefer something like:
```js
if (!showDiv) {
  props.id = id;
}
```
?

prateekbh:
>```
if (!showDiv) {
	props.id = id;
}
```

☝️ is preferred.

yeah, `var canUseDOM = !!(` can actually be better

**Comment by filoozom** [comment]
Should we use it then? Or should we keep the state for now and maybe move to that one later, maybe after testing is done to ensure that everything works well?

**Comment by prateekbh** [comment]
we should keep things as is now and come back later

**Comment by filoozom** [comment]
Changed a few things again, hope I didn't messed up and that it'll do it 😛 Found a small bug where it added ```aria-controls``` without an helper text, so removed that.
I feel like ```const showDiv = props.helptext || (props.label && !showFloatingLabel);``` was changed to ```const showDiv = props.helptext || props.label & !showFloatingLabel;``` automatically which didn't work, or maybe I need a lot more sleep than I thought I did...

File: Textfield/Textfield.jsx [hunk 8]
```diff
@@ -1,51 +1,175 @@
...
+    this.MDComponent && this.MDComponent.destroy && this.MDComponent.destroy();
   }
   materialDom(allprops) {
-    const { className, ...props } = allprops;
-    let labelClass = ["mdc-textfield__label"];
-    if (props.value) {
-      labelClass.push("mdc-textfield__label--float-above");
+    let { className, ...props } = allprops;
+
+    if (props.value && this.state.showFloatingLabel) {
```
prateekbh: change this to 'value' in props. that will fix the null state problem

PR #196, opened 2017-07-29
Add menu prop to Toolbar

plusCubed:
From the [MDC Toolbar docs](https://material.io/components/web/catalog/toolbar):
> There are two types of icons, `mdc-toolbar__icon--menu` represents the left most icon in `mdc-toolbar` usually to the left of `mdc-toolbar__title`

This will allow fixing the left/right padding on toolbar icons. I'm working on updating the demo/docs site. (AFAIK a release has to be made before I can compile the docs site?)

### DIFF
--- a/Toolbar/Toolbar.jsx
+++ b/Toolbar/Toolbar.jsx
@@ -65,6 +65,7 @@ class ToolbarIcon extends MaterialComponent {
   constructor() {
     super();
     this.componentName = "toolbar__icon";
+    this._mdcProps = ["menu"];
   }
   materialDom(props) {
     return (
--- a/Toolbar/Toolbar.jsx
+++ b/Toolbar/Toolbar.jsx
@@ -62,10 +62,10 @@ class ToolbarSection extends MaterialComponent {
 }
 
 class ToolbarIcon extends MaterialComponent {
-  constructor() {
+  constructor(props) {
     super();
     this.componentName = "toolbar__icon";
-    if (this.props.menu) {
+    if (props.menu) {
       this.componentName += "--menu";
     }
   }
--- a/Toolbar/Toolbar.jsx
+++ b/Toolbar/Toolbar.jsx
@@ -61,6 +61,9 @@ class ToolbarSection extends MaterialComponent {
   }
 }
 
+/**
+ * @prop menu = false
+ */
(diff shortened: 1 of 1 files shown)


filoozom:
Seems to work, but as we'd like to have the exact same output as ```material-components-web``` we would prefer to not have ```mdc-toolbar__icon``` when ```mdc-toolbar__icon--menu``` is used.
So maybe it would be better to make a ```Toolbar.MenuIcon``` component (which seems kind of overkill) or to set the ```className``` based on whether or not the ```menu``` prop is used.

plusCubed:
Ah, so with the current way, it would generate both class names. I'll implement the second solution (for now) - unless there's precedent in another component for this situation?

prateekbh:
I guess a new component would be great, check if u can figure a way to use `icon`. Also please check if mdc-web allows to change the icon itself from the hamburger menu. In this case lets take the content of icon from a prop or markup

plusCubed:
MDC-Web doesn't really limit what to put inside an element with that class. For what it's worth, in the [MDC demos for Toolbar](https://github.com/material-components/material-components-web/tree/master/demos/toolbar), they conform to how we're generating it right now. 

I don't think `Icon` would work as it uses `<i class="material-icons">`, while toolbar icons use `<a class="material-icons">` (at least in the mdc demos).

Toolbar/Toolbar.jsx
```diff
@@ -62,9 +62,12 @@ class ToolbarSection extends MaterialComponent {
 }
 
 class ToolbarIcon extends MaterialComponent {
-  constructor() {
+  constructor(props) {
     super();
     this.componentName = "toolbar__icon";
+    if (props.menu) {
```
prateekbh: y conditional?
isn't this have to be appended everytime?

plusCubed: The prop is used for the left-most icon. For other icons the `--menu` shouldn't be used. I didn't make a new component but rather reused the old one, just added a prop. If you want I could make a separate component instead (I think that goes against DRY though).

prateekbh: i guess approach is fine as it keeps minimal code on our side. 👍

PR #199, opened 2017-07-30
Various fixes and improvements for docs site

plusCubed:
Content:
- Update toolbar docs (#196)
- Fix wrong title on icon page
- Change drawer IconToggle title for consistency
- Add note and change sample to import List's css when using Menu and Select

Layout:
- Fix padding for site's toolbar (#196)
- Fix toolbar height on mobile devices
- Fix "Dark Mode" switch's theme
- Fix top padding for component page titles
- Use LayoutGrid for content - fixes right padding and mobile padding

Needs Review:
- Remove unused (?) sample folder: I believe the main one is updated on the master branch?
- Add max-width to central content: This looks better to me, but of course it's a style thing, so let me know if I should revert it. Ideally, it should use the permanent drawer on desktop, which would make it look less awkward, but that isn't fully implemented yet.


Components/Home/Home.css
```diff
@@ -27,8 +33,7 @@
 		margin: 0;
 	}
 	.content{
-		padding-left: 32px;
```
prateekbh: These padding are important for a clean negative margin

plusCubed: I don't see where negative margin is used? I replaced the manual padding with LayoutGrid, which has by default 24px padding on desktop and 16px padding on mobile. I can change it to 32px on desktop if you'd like that.

prateekbh:
Please upgrade to version 1.1.1

prateekbh:
wow the new docs looks great!!! 🎉 🎉 🎉 🎉 🎉 🎉

plusCubed:
Glad to be of help! This project has been awesome to use in my own, and much better than the other ones out there.

About the dark theme toolbar section - it was to make the switch light-colored and visible. I wasn't able to figure out what it was on a small screen the first time I visited the site. 

Also it looks like the new toolbar fix wasn't applied? I think maybe the site is built with the old preact-material-components - probably need to run `npm update`. This is how the toolbar spacing should look:
fixed toolbar

prateekbh:
have your changes made it to 1.1.1 ?

plusCubed:
Yes, they are in the compiled index.js for Toolbar.

prateekbh:
then a simple update should work

plusCubed:
Hmmm I just reinstalled 1.1.1 and it's working as the image above. Can you check node_modules/preact-material-components/package.json? 

(Also, I now see the problem with the dark theme causing a dark background - I'll make a new pull request for that.)

prateekbh:
i have fixed the menu for you 👍 

>(Also, I now see the problem with the dark theme causing a dark background - I'll make a new pull request for that.)

I didn't get what you mean here

PR #204, opened 2017-08-01
Fixes Update issues with Slider component. 

MobiusHorizons:
The Slider component was not updating when props changed. This update allows the component to be used properly in a controlled component manner.

### DIFF
--- a/Slider/Slider.jsx
+++ b/Slider/Slider.jsx
@@ -27,41 +27,39 @@ export default class Slider extends MaterialComponent {
     this.MDComponent = new MDCSlider(this.base);
     this.MDComponent.listen("MDCSlider:change", this._onChange);
     this.MDComponent.listen("MDCSlider:input", this._onInput);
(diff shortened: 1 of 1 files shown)


Slider/Slider.jsx
```diff
@@ -27,41 +27,39 @@ export default class Slider extends MaterialComponent {
...
-      value,
-      step,
-      tabindex = 0,
-      ...props
-    } = allprops;
+  setValue(props = this.props) {
+    const { disabled = false, min = 0, max = 100, value, step } = props;
+    if (this.MDComponent) {
+      if (min > this.MDComponent.max) {
+        this.MDComponent.max = max;
```
prateekbh: what am i missing?
both conditions seems to have same code

MobiusHorizons: max before min or min before max. it's because MDC throws an error if you set min to be greater than max.

PR #205, opened 2017-08-01
Add dark theme class to the toolbar

plusCubed:
Old:
untitled

New:
untitled2

This makes the switch more visible. (cont. from #199)


prateekbh:
A less bright white switch is more visible than the the older more whiter one????

plusCubed:
Mostly the right side which blends in, on the old one it looks more like just a white circle.

prateekbh:
@plusCubed can you send the screenshot of the entire page?

plusCubed:
Old:
prateekbh github io-preact-material-components-

New:

plusCubed:
The new one is designed to be used on dark backgrounds by MDC.

PR #208, opened 2017-08-03
Improve organization and simplify compilation

plusCubed:
- Remove `compileComponents.js`, instead:
    - Run babel-cli to compile `jsx` to `js`
         - Use transform-runtime for smaller ES6 exports
    - Include `index.js` to export each component `js` file
- Use webpack solely for UMD bundle
     - Revert back to simple `copyCss.js` script
- Add .gitattributes for line ending normalization


package.json
```diff
@@ -5,10 +5,13 @@
   "module": "index.js",
   "main": "dist/index.js",
   "scripts": {
-    "build": "node ./compileComponents.js && webpack",
-    "prepublish": "npm run build && npm run test ",
+    "build": "npm run build:umd && npm run build:es",
+    "build:umd": "cross-env NODE_ENV=production webpack --config webpack.config.umd.js",
+    "build:es": "cross-env NODE_ENV=production babel . --out-dir . --extensions \".jsx\" --ignore node_modules,sample && npm run copycss",
+    "copycss": "node copyCss.js",
```
prateekbh: why a separate command?

copyCss.js
```diff
@@ -0,0 +1,56 @@
+/*globals require, __dirname*/
```
prateekbh: why the rename?

webpack.config.umd.js
```diff
@@ -10,26 +8,30 @@ module.exports = {
...
   module: {
     rules: [
       {
         loader: "babel-loader",
         test: /\.(js|jsx)$/,
         exclude: /node_modules\/proptypes|scripts\/sw.js/,
         options: {
-          presets: [["es2015", { modules: false }]],
+          babelrc: false,
+          presets: ["es2015", "react", "stage-0"],
```
prateekbh: wont this convert the entire code into ES5?

webpack.config.umd.js
```diff
@@ -10,26 +8,30 @@ module.exports = {
...
+  },
   module: {
     rules: [
       {
         loader: "babel-loader",
         test: /\.(js|jsx)$/,
         exclude: /node_modules\/proptypes|scripts\/sw.js/,
         options: {
-          presets: [["es2015", { modules: false }]],
+          babelrc: false,
```
prateekbh: correct me if I am wrong, wont keeping this true, use our bablerc as a single source of truth?

webpack.config.umd.js
```diff
@@ -10,26 +8,30 @@ module.exports = {
...
-            ["transform-react-jsx", { pragma: "h" }],
-            "transform-object-rest-spread"
+            ["transform-react-jsx", { pragma: "h" }]
           ]
         }
       }
     ]
   },
   plugins: [
-    new CssMigrationWebpackPlugin(),
```
prateekbh: same question... why remove this as a part of single process?

package.json
plusCubed: For ES6, we weren't using Webpack anyway, just a babel command to compile the jsx and earlier-stage-ES6. This makes it clearer and puts it into one command. I'm somewhat copying what Brian is doing with react-virtualized [here](https://github.com/bvaughn/react-virtualized/blob/master/package.json).

copyCss.js
plusCubed: This file only copies css instead of compiling.

webpack.config.umd.js
plusCubed: Webpack is only used for UMD/ES5 bundling (the dist bundle)

webpack.config.umd.js
plusCubed: I agree that this is ugly, maybe there's a more elegant way to do this by just adding the es2015 preset onto babelrc's config...?

webpack.config.umd.js
plusCubed: Separation of concerns into 3 parts:
- Webpack for UMD bundling
- Babel command for ES6
- CSS copying, as it has nothing to with JS code compiling

prateekbh: actually repo had 3 different commands to all this but people felt that one single command should just build the entire thing, hence they were all combined under one command

plusCubed: It's still under one command, the `npm run build` still works like before. I'm modeling these build steps from the [react-virtualized](https://github.com/bvaughn/react-virtualized/blob/master/package.json) project

prateekbh:
changes look fine, but can you send the PR to `orgImprovement` branch instead? would like to try before merging.

prateekbh:
@plusCubed sorry for the delay, The setup looks good. Except if you can import components from their `.js` instead of creating a new index.js.
Also please I guess no extra options for UglifyPlugin would be better.

After its good to be merged with master 👍

copyCss.js
```diff
@@ -0,0 +1,56 @@
...
+        srcPath,
+        source,
+        "dist",
+        "mdc." + source + ".css"
+      );
+      const destFolderPath = path.join(__dirname, dest);
+
+      //create folder if not already present
+      if (!fs.existsSync(destFolderPath)) {
+        mkdirp(destFolderPath);
```
filoozom: Don't forget to change it to ```mkdirp.sync``` as per #234.

package.json
```diff
@@ -5,10 +5,13 @@
...
+    "prepublish": "npm run test",
```
filoozom: This should also be renamed to ```prepublishOnly```.

PR #209, opened 2017-08-03
Add dialog accept and cancel events

plusCubed:
https://github.com/material-components/material-components-web/tree/master/packages/mdc-dialog#dialog-events

### DIFF
--- a/Dialog/Dialog.jsx
+++ b/Dialog/Dialog.jsx
@@ -8,6 +8,8 @@ class Dialog extends MaterialComponent {
   constructor() {
     super();
     this.componentName = "dialog";
+    this._onAccept = this._onAccept.bind(this);
+    this._onCancel = this._onCancel.bind(this);
   }
   componentDidMount() {
     this.MDComponent = new MDCDialog(this.control);
--- a/Dialog/Dialog.jsx
+++ b/Dialog/Dialog.jsx
@@ -11,9 +11,23 @@ class Dialog extends MaterialComponent {
   }
   componentDidMount() {
     this.MDComponent = new MDCDialog(this.control);
+    this.MDComponent.listen("MDCDialog:accept", this._onAccept);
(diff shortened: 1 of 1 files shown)


Dialog/Dialog.jsx
```diff
@@ -11,10 +11,24 @@ class Dialog extends MaterialComponent {
   }
   componentDidMount() {
     this.MDComponent = new MDCDialog(this.control);
+    this.MDComponent.listen("MDCDialog:accept", this._onAccept);
```
prateekbh: you need to to bind it with `this` in the constructor

Dialog/Dialog.jsx
```diff
@@ -8,13 +8,29 @@ class Dialog extends MaterialComponent {
...
+    this._onAccept = this._onAccept.bind(this);
+    this._onCancel = this._onCancel.bind(this);
   }
   componentDidMount() {
     this.MDComponent = new MDCDialog(this.control);
+    this.MDComponent.listen("MDCDialog:accept", this._onAccept);
+    this.MDComponent.listen("MDCDialog:cancel", this._onCancel);
   }
   componentWillUnmount() {
+    this.MDComponent.unlisten("MDCDialog:accept", this._onAccept);
```
prateekbh: `this.MDComponent.unlisten && this.MDComponent.unlisten(...`

plusCubed: The method will always exist, no? I mean we could check `this.MDComponent` instead, if that's necessary

prateekbh: no so its not a public method so far, hence the check

plusCubed: https://material.io/components/web/catalog/base/#methods?

prateekbh: oh sry my bad, got confused

PR #217, opened 2017-08-10
Fix MaterialComponent className

### LINKED ISSUE #213: Missing css class when build
Run fine when use command preact watch. But when use preact build it miss the class ```"mdc-temporary-drawer mdc-typography"``` and ```"material-icons mdc-toolbar__icon--menu"```. So for the UI to work, i must add thoes class manually like this:
```
<Toolbar>
  <Toolbar.Row>
    <Toolbar.Section align-start>
      <Toolbar.Icon className="material-icons mdc-toolbar__icon--menu" menu href="#" onClick={this.openDrawer}>menu</Toolbar.Icon>
	<Toolbar.Title>
	  Joider
       </Toolbar.Title>
       </Toolbar.Section>
 </Toolbar.Row>
</Toolbar>
<Drawer.TemporaryDrawer className="mdc-temporary-drawer mdc-typography" ref={drawer => { this.drawer = drawer; }} >
  <Drawer.TemporaryDrawerContent>
    <List>
     <List.LinkItem>
	<List.ItemIcon>home</List.ItemIcon>
          Home
	</List.LinkItem>
    </List>
  </Drawer.TemporaryDrawerContent>
</Drawer.TemporaryDrawer>
```

filoozom:
Fixes #213

### DIFF
--- a/MaterialComponent.js
+++ b/MaterialComponent.js
@@ -61,9 +61,8 @@ export default class MaterialComponent extends Component {
     // Fetch a VNode
     const element = this.materialDom(this.props);
     element.attributes = element.attributes || {};
-    const autoInits = this.getAutoInitNames();
     // Fix for className
-    element.attributes.class = this.getClassName(element);
+    element.attributes.className = this.getClassName(element);
     // Clean this shit of proxy attributes
     this._mdcProps.forEach(prop => {
       delete element.attributes[prop];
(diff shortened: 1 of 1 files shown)


plusCubed:
This change broke CSS when there is already a className on the component.

For example, `<Toolbar className="mdc-theme--dark">` only has `mdc-theme--dark` classes in the rendered HTML instead of `mdc-theme--dark mdc-toolbar`.

I think this is because Preact [duplicates the class and className attributes](https://github.com/developit/preact/blob/33fc697ac11762a1cb6e71e9847670d047af7ce5/src/dom/index.js#L36), so we'd have to do that here too (re-add the original line & keep both lines)

PR #237, opened 2017-09-02
[WIP] Add TextfieldInput ref to Textfield and fix focused state

### LINKED ISSUE #236: Textfield layout breaks when no value
If you have a Textfield and delete it's value, the layout can become broken:

textfield

It looks like because it only has this class when there is a value: https://github.com/prateekbh/preact-material-components/blob/master/Textfield/Textfield.jsx#L73-L75

This can occur when your Textfield is connected to state, e.g,:
```jsx
<Textfield value={ this.state.lastName } onInput={ e => this.setState({ lastName: e.target.value }) } />
```

filoozom:
Closes #236.

- Adds ```input``` as a reference to the underlying ```TextfieldInput``` of a ```Textfield```. Makes it possible to access ```MDComponent```: ```textfieldRef.input.MDComponent```.
- Keeps floating label when input is focused but value is falsy.

To do:
- Test if this doesn't break anything
- Keep ```mdc-textfield--upgraded``` on JS-enabled clients because labels currently don't work correctly after the first focus.

### DIFF
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -87,10 +87,16 @@ class TextfieldInput extends MaterialComponent {
 
   materialDom(allprops, { showFloatingLabel, isFocused }) {
     let { className, ...props } = allprops;
+    let labelClassName = "mdc-textfield__label";
+
     const upgrade = props.value || isFocused;
 
-    if (showFloatingLabel && upgrade) {
+    if (showFloatingLabel) {
       className = [className, "mdc-textfield--upgraded"].join(" ");
+    }
+
+    if (upgrade) {
+      labelClassName = [labelClassName, "mdc-textfield__label--float-above"].join(" ");
     }
 
     return (
@@ -113,7 +119,7 @@ class TextfieldInput extends MaterialComponent {
           this.state.showFloatingLabel &&
           <Label
             for={props.id}
-            className={upgrade && "mdc-textfield__label--float-above"}
+            className={labelClassName}
           >
             {props.label}
           </Label>}
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -51,9 +51,13 @@ class TextfieldInput extends MaterialComponent {
     this.componentName = "textfield";
     this._mdcProps = ["fullwidth", "multiline", "dense", "disabled", "box"];
     this.state = {
-      showFloatingLabel: false
+      showFloatingLabel: false,
+      isFocused: false
     };
+    this._onFocus = this._onFocus.bind(this);
+    this._onBlur = this._onBlur.bind(this);
   }
+
   componentDidMount() {
     this.setState(
       {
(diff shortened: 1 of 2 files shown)


prateekbh:
sorry, jumping in on your WIP PR.  just trying to avoid any extra work

Textfield/Textfield.jsx
```diff
@@ -64,33 +68,58 @@ class TextfieldInput extends MaterialComponent {
...
+    this.setState({
+      isFocused: false
+    });
+  }
+
+  materialDom(allprops, { showFloatingLabel, isFocused }) {
     let { className, ...props } = allprops;
+    let labelClassName = "mdc-textfield__label";
+
+    const upgrade = props.value || isFocused;
```
prateekbh: so upgrade is not the function of value or focus, its just if the js enhancement has been done or not

Textfield/Textfield.jsx
```diff
@@ -64,33 +68,58 @@ class TextfieldInput extends MaterialComponent {
...
 
-    if (props.value && this.state.showFloatingLabel) {
+    if (showFloatingLabel) {
       className = [className, "mdc-textfield--upgraded"].join(" ");
```
prateekbh: sorry I might have forgot, y are we doing this by hand when `mdc-web` is already doing this for us?

Textfield/Textfield.jsx
```diff
@@ -64,33 +68,58 @@ class TextfieldInput extends MaterialComponent {
...
               className="mdc-textfield__input"
+              onFocus={this._onFocus}
+              onBlur={this._onBlur}
               {...props}
             />}
         {props.label &&
           this.state.showFloatingLabel &&
           <Label
             for={props.id}
-            className={props.value && "mdc-textfield__label--float-above"}
```
prateekbh: `props.value` check is needed cuz it floats above only when some value is there. float on focus is auto handled by `mdc-web`

Textfield/Textfield.jsx
```diff
@@ -134,6 +163,13 @@ class Textfield extends Component {
     return ++this.uidCounter;
   }
 
+  _setInputRef(ref) {
+    if (ref) {
+      this.input = ref;
```
prateekbh: what are we trying to do with ref of input or mdcomponent?

Textfield/Textfield.jsx
```diff
@@ -64,33 +68,58 @@ class TextfieldInput extends MaterialComponent {
...
+
+    const upgrade = props.value || isFocused;
 
-    if (props.value && this.state.showFloatingLabel) {
+    if (showFloatingLabel) {
       className = [className, "mdc-textfield--upgraded"].join(" ");
     }
 
+    if (upgrade) {
+      labelClassName = [labelClassName, "mdc-textfield__label--float-above"].join(" ");
```
prateekbh: this is well handled by `mdc-web` just the `'value' in props` change would fix the bug in hand

Textfield/Textfield.jsx
```diff
@@ -64,33 +68,58 @@ class TextfieldInput extends MaterialComponent {
...
+  }
+
+  materialDom(allprops, { showFloatingLabel, isFocused }) {
     let { className, ...props } = allprops;
+    let labelClassName = "mdc-textfield__label";
+
+    const upgrade = props.value || isFocused;
 
-    if (props.value && this.state.showFloatingLabel) {
+    if (showFloatingLabel) {
```
prateekbh: let this condition be 
`if ('value' in props && this.state.showFloatingLabel)` this will handle the null state problem

prateekbh:
closing as per conversation

PR #238, opened 2017-09-04
label bug fix

### LINKED ISSUE #236: Textfield layout breaks when no value

prateekbh:
fixes #236

### DIFF
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -70,17 +70,14 @@ class TextfieldInput extends MaterialComponent {
   materialDom(allprops) {
     let { className, ...props } = allprops;
 
-    if (props.value && this.state.showFloatingLabel) {
+    if ("value" in props && this.state.showFloatingLabel) {
       className = [className, "mdc-textfield--upgraded"].join(" ");
     }
 
     return (
       <div className={className} ref={control => (this.control = control)}>
         {props.multiline
-          ? <textarea
-              className="mdc-textfield__input"
-              {...props}
-            />
+          ? <textarea className="mdc-textfield__input" {...props} />
           : <input
               type={props.type || "text"}
               className="mdc-textfield__input"
@@ -160,7 +157,10 @@ class Textfield extends Component {
             <label for={props.id}>
               {props.cssLabel || `${props.label}: `}
             </label>}
-          <TextfieldInput {...props} aria-controls={props.helptext && props.id + "-helptext"} />
+          <TextfieldInput
+            {...props}
+            aria-controls={props.helptext && props.id + "-helptext"}
+          />
           {props.helptext &&
             <Helptext id={props.id + "-helptext"} {...helptextProps}>
               {props.helptext}
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -85,10 +85,7 @@ class TextfieldInput extends MaterialComponent {
             />}
         {props.label &&
           this.state.showFloatingLabel &&
-          <Label
-            for={props.id}
-            className={props.value && "mdc-textfield__label--float-above"}
-          >
+          <Label for={props.id}>
             {props.label}
           </Label>}
       </div>
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -85,14 +85,7 @@ class TextfieldInput extends MaterialComponent {
             />}
         {props.label &&
           this.state.showFloatingLabel &&
-          <Label
-            for={props.id}
-            className={
-              this.state.showFloatingLabel
-                ? ""
-                : "mdc-textfield__label--float-above"
-            }
-          >
+          <Label for={props.id}>
             {props.label}
           </Label>}
       </div>
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -85,7 +85,14 @@ class TextfieldInput extends MaterialComponent {
             />}
         {props.label &&
           this.state.showFloatingLabel &&
-          <Label for={props.id}>
+          <Label
+            for={props.id}
+            className={
+              this.state.showFloatingLabel
+                ? ""
+                : "mdc-textfield__label--float-above"
+            }
+          >
             {props.label}
           </Label>}
       </div>


prateekbh:
no actually that'll need to be truthy and not just present... 
`'value' in props` @ the place u r suggesting will make the label float even for a null or empty value

filoozom:
That's what I though too, but when I checkout your PR it doesn't work as expected on my test page while it does work with ```'value' in props``` on the label.

I'm testing with:
```jsx
export default class Home extends Component {

	state = {
		value: ''
	}

	render() {
		return (
			<div class={style.home}>
				<h1>Home</h1>
				<p>This is the Home component.</p>
				<Textfield id="test" />
				<Textfield label="lala" id="test2" />
				<Textfield label="No ID" />

				<Textfield label="State" value={this.state.value} onInput={e => this.setState({ value: e.target.value })} ref={ref => window.mdcRef = ref} required id="state" />
			</div>
		);
	}
}
```

Does your PR work when you write something in the input and then remove everything from it?

prateekbh:
yep, lemme check again though.

filoozom:
>'value' in props @ the place u r suggesting will make the label float even for a null or empty value

It should actually, shouldn't it? As long as it's focused of course.

prateekbh:
so i checked, its working fine for me though...
also the focus thing is handled by `mdc-web` so we dont need to do anything extra

filoozom:
Strange. @gpoitch can you check if this fixes your issues?

prateekbh:
by the way @filoozom what happens when you remove everything from the Textfield?

filoozom:
The label teleports down as a "placeholder" instead of still floating while it's focused. I'll try to make a gif.

prateekbh:
@filoozom make sure ur changes are in textfield's index.js too 😛

filoozom:
```bash
$ git fetch upstream pull/238/head:pr-fix-236
remote: Counting objects: 4, done.
remote: Total 4 (delta 2), reused 3 (delta 2), pack-reused 0
Unpacking objects: 100% (4/4), done.
From https://github.com/prateekbh/preact-material-components
 * [new ref]         refs/pull/238/head -> pr-fix-236
$ git checkout pr-fix-236
Switched to branch 'pr-fix-236'
$ git show HEAD
commit e5184143c5cb9f5263d8c968e716df7c6110d959 (HEAD -> pr-fix-236)
Author: Prateek Bhatnagar <prateek89born@gmail.com>
Date:   Mon Sep 4 15:28:38 2017 +0530

    label bug fix

diff --git a/Textfield/Textfield.jsx b/Textfield/Textfield.jsx
index e514078..ab9b4b8 100644
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -70,17 +70,14 @@ class TextfieldInput extends MaterialComponent {
   materialDom(allprops) {

-    if (props.value && this.state.showFloatingLabel) {
+    if ("value" in props && this.state.showFloatingLabel) {
       className = [className, "mdc-textfield--upgraded"].join(" ");
     }

     return (
       <div className={className} ref={control => (this.control = control)}>
         {props.multiline
-          ? <textarea
-              className="mdc-textfield__input"
-              {...props}
-            />
+          ? <textarea className="mdc-textfield__input" {...props} />
           : <input
               type={props.type || "text"}
               className="mdc-textfield__input"
@@ -160,7 +157,10 @@ class Textfield extends Component {
             <label for={props.id}>
               {props.cssLabel || `${props.label}: `}
             </label>}
-          <TextfieldInput {...props} aria-controls={props.helptext && props.id + "-helptext"} />
+          <TextfieldInput
+            {...props}
+            aria-controls={props.helptext && props.id + "-helptext"}
+          />
           {props.helptext &&
             <Helptext id={props.id + "-helptext"} {...helptextProps}>
               {props.helptext}
$ npm run build

> preact-material-components@1.1.6 build /Users/filoozom/Projects/preact-mdc/node_modules/preact-material-components
> node ./compileComponents.js && webpack

Executing post-build scripts
Hash: be208bc15416bfac7a2f
Version: webpack 3.5.2
Time: 4728ms
   Asset    Size  Chunks                    Chunk Names
index.js  628 kB       0  [emitted]  [big]  main
   [8] ./List/index.js 9.96 kB {0} [built]
  [10] ./Button/index.js 3.54 kB {0} [built]
  [21] ./index.js 917 bytes {0} [built]
  [49] ./Select/index.js 4.42 kB {0} [built]
  [57] ./Card/index.js 7.79 kB {0} [built]
  [58] ./Elevation/index.js 2.88 kB {0} [built]
  [59] ./Formfield/index.js 1.48 kB {0} [built]
  [60] ./Dialog/index.js 6.49 kB {0} [built]
  [67] ./Drawer/index.js 12.7 kB {0} [built]
  [76] ./Toolbar/index.js 6.05 kB {0} [built]
  [81] ./LinearProgress/index.js 3.53 kB {0} [built]
  [85] ./LayoutGrid/index.js 6.38 kB {0} [built]
  [86] ./Menu/index.js 3.59 kB {0} [built]
  [87] ./Tabs/index.js 6.77 kB {0} [built]
  [95] ./Slider/index.js 5.36 kB {0} [built]
    + 83 hidden modules
```

Then of course rebuilding the test app with ```preact-cli```, then tested. When I focus the first time it works as expected, then I enter one letter, remove it, and the label floats down to its empty non-focused position.

Can't make a gif right now but will try to do it as soon as possible.

prateekbh:
in the `textfield` folder there must be an `index.js` check if same changes are there too or not

filoozom:
screen shot 2017-09-04 at 17 02 30

as opposed to

screen shot 2017-09-04 at 17 03 41

which works as expected.
Same sample code as in https://github.com/prateekbh/preact-material-components/pull/238#issuecomment-326925240.

prateekbh:
cool will take a look at this in the morning and send the fix over

prateekbh:
checked it, seems like we can completely get rid of props.value for `mdc-textfield__label--float-above`. `mdc-web` takes care of it and any such custom logic like this(sadly I introduced it) takes care of it.

prateekbh:
hmmm makes sense, how about we keep this and remove it after `componentDidMount` ?

gpoitch:
Somewhat unrelated, but attributes such as `multiline, label, helptext, dense`, etc should be destructured off of `allprops` otherwise those (invalid) html attributes get set on the raw DOM elements.

Textfield/Textfield.jsx
```diff
@@ -85,7 +85,14 @@ class TextfieldInput extends MaterialComponent {
             />}
         {props.label &&
           this.state.showFloatingLabel &&
-          <Label for={props.id}>
+          <Label
+            for={props.id}
+            className={
+              this.state.showFloatingLabel
```
filoozom: If I'm not missing something, this condition is always true so this is actually exactly the same code as before this commit.

prateekbh: even I thought so, so this will not be true in case of SSR(cDM works on client only r8?), hence an SSR'd html will have this class and avoid FOUC

filoozom:
@gpoitch Those attributes should be removed here:
https://github.com/prateekbh/preact-material-components/blob/2ac3e39de474f028cf3fc76feebcd40af3bfe6c2/MaterialComponent.js#L65-L67

Have you had any issues with attributes being added to the DOM? If so, please open a new issue.

Textfield/Textfield.jsx
filoozom: Yes but the condition to show the label is ```props.label && this.state.showFloatingLabel```, so in this case ```this.state.showFloatingLabel``` must be true, right?
In case of SSR we show the other label, the one in ```Textfield``` and not ```TextfieldInput```. Maybe this means we can't avoid FOUC?

gpoitch:
@filoozom I can see the attributes on the demo site.  Not causing any bugs, but I still don't think they should be there (performance / valid html).  I'll open a separate issue

Textfield/Textfield.jsx
prateekbh: >Yes but the condition to show the label is props.label && this.state.showFloatingLabel, so in this case this.state.showFloatingLabel must be true, right?

Oh yes silly me.

lets revert my last commit and close publish this, also lets open another issue for FOUC

PR #242, opened 2017-09-06
adding 0.20.0

prateekbh:
new Stroked Buttons
and MDComponent for Textfield


.gitignore
```diff
@@ -46,3 +46,6 @@ jspm_packages
 # Ignores all index.js files for all components since they are auto-generated.
 **/index.js
 !/index.js
+
+
+package-lock.json
```
MobiusHorizons: You probably have a reason for ignoring this file, but I think you are supposed to commit it to source control so that updates to dependencies only happen explicitly when you update them.

sample/home.jsx
```diff
@@ -80,6 +80,7 @@ export default class Home extends Component {
           ripple={true}
           onClick={e => {
             this.drawer.MDComponent.open = true;
+            console.log(this.tb.MDComponent);
```
filoozom: Is this necessary?

sample/home.jsx
```diff
@@ -144,7 +145,14 @@ export default class Home extends Component {
           <label for="rd">Radio Box</label>
         </div>
 
-        <Textfield label="hello" fullwidth={true} required />
+        <Textfield
+          label="hello"
+          fullwidth
+          required
+          ref={tb => {
```
filoozom: Same here, it's interesting for testing but maybe shouldn't be in the sample. Should definitely update the docs though, which will be much easier to enforce once the docs source code is on ```master``` because we can just refuse PRs if they don't update docs 👍

.gitignore
filoozom: Yes, this should indeed be committed. Could avoid potential issues with libraries that don't respect semver.

PR #244, opened 2017-09-07
updating documentation

prateekbh:
updating documentation for 1.1.6


prateekbh:
also , I need to check how to remove travis from this branch

filoozom:
I'll be doing some work on Travis for the documentation update to ```preact-cli```, will have a look at the same time.

prateekbh:
will do

prateekbh:
done

prateekbh:
I guess we'll need to force submit this one

filoozom:
Yes, tried to re-run the build to see if it would just ignore the build and accept it, but it didn't want to. Could also just add a commit, should do the trick I think.

prateekbh:
about package.json, i dont really have a strong opinion, but yeah i guess I should make peace with it... will add it in next pushes

filoozom:
Wait, why did you add ```espree``` and ```babel-generator```?

prateekbh:
`babel-generator` was due to the error being thrown and espree also i guess was a sub dependency

PR #245, opened 2017-09-07
Enforce code style

### LINKED ISSUE #243: Enforce code style
We should enforce a given code style and improve linting for the source code. For now, a lot of PRs contain lines that don't change in the slightest except their formatting, which makes it more difficult to track changes and review pull requests.

Also, I think that the current code doesn't respect all ```eslint``` rules. At least VS Code shows errors, I'll have to check if they're indeed from our eslint config.

filoozom:
This does following things:
- Sync ```eslint``` and ```prettier``` configs. This means that linting won't show errors for code fixed by ```prettier```.
- Make ```prettier``` mandatory, else travis builds will fail (added to ```test``` script)
- Run ```prettier``` on all files

Closes #243.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,76 +1,22 @@
 parser: babel-eslint
-extends: eslint:recommended
+
+extends:
+  - prettier
+  - prettier/react
+
+plugins:
+  - react
+  - prettier
+
 parserOptions:
-  ecmaVersion: 2015
+  ecmaVersion: 2016
+  sourceType: module
   ecmaFeatures:
-    modules: true
     jsx: true
 
 env:
- browser: true
- mocha: true
-globals:
-  sinon: true
-  expect: true
-
-plugins:
- - react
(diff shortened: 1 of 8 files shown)


prateekbh:
just curious, should we enforce single quotes? looks prettier :p

filoozom:
I usually do, but this project contains a lot of double quotes so I didn't. Will gladly add it though 👍.
Damn, that was a real quick review... 😛

prateekbh:
Lol I get notifications for evverything

filoozom:
Actually ```prettier``` prefers double quotes, and we should use them for JSX, so maybe we should keep it like that.

prateekbh:
cool makes sense

filoozom:
Guess there's no need to publish a new version for this as it actually doesn't change anything except for contributors, so I can just go ahead and merge this, right?

prateekbh:
yep

PR #221, opened 2017-08-14
[WIP] Port gh-pages to preact-cli

filoozom:
This should make ```gh-pages``` more easy to maintain because we wouldn't need to use ```webpack``` / ```babel``` etc anymore. It also aims to use technologies that might be more know to ```preact``` developers (```preact-cli``` and ```preact-router``` instead of ```routie```).

- [x] Have same functionality as currenctly (except it can't be published)
- [x] Replace ```PostCSS``` with ```Sass``` to make it compatible with ```preact-cli```
- [x] Clean old build code
- [x] Replace ```routie``` with ```preact-router``` (required for SSR support as far as I can tell)
- [x] Enable SSR for all pages
- [x] Rename all routes (components) to the same names as ```material-components-web```
- [x] Comment dependencies (makes code a little cleaner I think)
- [ ] Use CSS Modules (everything is ```:global``` for now)
- [ ] Publish gh-pages using ```preact build``` and deploy the ```build``` directory (how could this be done? new branch?)
- [ ] Lint code
- [ ] Provide custom ```template.html```

By the way, this also fixes #220.

PR #246, opened 2017-09-07
Add docs to master and set up automatic gh-pages update

filoozom:
This is a follow-up of #221.

Not quite sure how the automatic deploy to ```gh-pages``` with Travis works for now, but I'm testing.


docs/.eslintrc
```diff
@@ -0,0 +1,65 @@
+{
```
prateekbh: shouldn't this be same as prettier one?

docs/.gitignore
```diff
@@ -0,0 +1,43 @@
+# Logs
```
prateekbh: does this folder need a separate gitignore?

docs/.eslintrc
filoozom: Yes it should, this is just the result of #221 moved to the ```docs``` folder on ```master```, I'm not quite done yet.

I think I'll just remove this .eslintrc file though, as it can be handled by the root one.

docs/.gitignore
filoozom: Nope, just /build.

filoozom:
@prateekbh Just opened this PR to test the Travis config, no need to review right now, I'll ping you once I'm done 😄.

filoozom:
Okay, I think that this basically works, just need to create a dummy user and add the personnal access token to Travis.

@prateekbh I think you can start to review now!

docs/.gitignore
filoozom: Kept a ```.gitignore``` in ```docs``` so that the root package is completely separated from the ```docs``` package, except for ```.travis.yml```. Same reason why I have ```prettier``` and ```eslint``` duplicated in both.

docs/.eslintrc.yml
```diff
@@ -0,0 +1,22 @@
+parser: babel-eslint
```
prateekbh: cant we just have parent folder's eslintrc?

docs/prerender-urls.json
```diff
@@ -0,0 +1,70 @@
+[{
```
prateekbh: keep this

docs/.eslintrc.yml
filoozom: Sure, I was just trying to have two different projects so that they are not depending. We can just move ```docs``` somewhere else and it would still work. Also, we can have different ```eslint``` rules for docs and the library.

docs/src/components/app/index.js
```diff
@@ -0,0 +1,344 @@
...
+
+// Class
+export default class Home extends Component {
+  constructor() {
+    super();
+    this.state = {
+      darkMode: false
+    };
+    this.menuItems = [
+      {
```
prateekbh: how about this, we add this info in `prerender.json` as a separate node, and import that json in this file and loop over it. will our code pretty cleaner

filoozom:
By the way, ```docs/src``` is just a copy of #221 so you can check there to see a cleaner diff.

docs/src/components/app/index.js
filoozom: Will check if that's doable. Can't quite have something like ```component: HomePage``` in JSON.

docs/.eslintrc.yml
filoozom: Should I remove ```prettier``` from ```docs``` too then?

docs/src/components/app/index.js
prateekbh: ahhh ok

docs/.eslintrc.yml
filoozom: @prateekbh Just this last thing to clarify and we can merge.

filoozom:
@prateekbh Don't forget to create the GitHub user and to add the token to travis before merging ;-)

Issue #240, opened 2017-09-05
Material specific props not being removed before render


gpoitch:
Internal mdc props are being set on raw DOM elements (e.g. ripple, label, helptext, etc).  Although this doesn't cause any bugs at the moment, these are not valid html attributes and have some overhead of having to be set.

From the demo site:
![screen shot 2017-09-05 at 11 27 09 am](https://user-images.githubusercontent.com/411908/30069624-3dfe0f6e-922f-11e7-84f7-c01116ad7cb6.png)

This code is supposed to account for it:
https://github.com/prateekbh/preact-material-components/blob/2ac3e39de474f028cf3fc76feebcd40af3bfe6c2/MaterialComponent.js#L65-L67

However, I think this should happen before `this.materialDom(this.props)` is called.

filoozom:
Oh, actually this code just removes mdcProps, not custom attributes we add to the components. Misunderstood what you meant on the other PR, guess we should indeed fix this. Should be able to add an array to every components and remove all contained attributes in ```MaterialComponent.js```. Or deconstruct every prop in ```materialDom```, might be better. What do you think @prateekbh?

prateekbh:
`MaterialComponent.js` seems to be the best place to do this

PR #247, opened 2017-09-07
Initial GridList implementation

edhille:
This is by no means final, but I wanted to get it in front of you for feedback.  A few things I should call out.

I used http://www.material-ui.com/#/components/grid-list as a guide for what properties to expose. Also, I went the route of using a div in place of an img per the "Using a div in place of an img" section of https://github.com/material-components/material-components-web/tree/master/packages/mdc-grid-list. The rationale here was that it is more flexible and the alternative of having some sort of "supportMixedImgSizes" property seemed a little awkward. That said, I'm completely amenable to heading in that direction if you'd prefer.

I tried to mirror the conventions I saw in the codebase and basically adapted the GridLayout for this component, but let me know if I diverged from the preferred style in any way.

If you think this is heading in the right direction, I'll finish adding support for the rest of the props and create a sample usage.

### DIFF
--- a/GridList/GridList.jsx
+++ b/GridList/GridList.jsx
@@ -1,178 +1,138 @@
 import { h } from "preact";
 import MaterialComponent from "../MaterialComponent";
+import Icon from "../Icon";
 
 /**
- * @prop cellHeight	{number} - Number of px for one cell height. You can set 'auto' if you want to let the children determine the height.
- * @prop cols {number} - Number of columns.
- * @prop padding {1|4} - Number of px for the padding/spacing between items.
- * @prop style {object} - Override the inline styles of the root element
+ * @prop tile-gutter {1|4} - Number of px for the padding/spacing between items.
+ * @prop header-caption {boolean} - position <GridList.Tile.Secondary> at top
+ * @prop twoline-caption {boolean} - add spacing to <GridList.Tile.Secondary> for <GridList.Tile.SupportText>
+ * @prop icon-align {"start"|"end"} - position <GridList.Tile.Icon> at beginning or end of <GridList.Tile.Secondary>
+ * @prop tile-aspect-ratio {"1x1"|"16x9"|"2x3"|"3x2"|"4x3"|"3x4"} - aspect ratio for <GridList.Tile.Primary>
  */
 class GridList extends MaterialComponent {
   constructor() {
     super();
     this.componentName = "grid-list";
-    this._propsDict = {
-      cols: "cols",
-      padding: "padding",
-      cellHeight: "cellHeight",
-      secondaryActionIconAlign: "secondaryActionIconAlign",
-      aspectRatio: "aspectRatio",
-      twoLineCaptions: "twoLineCaptions"
-    };
-  }
-  createClassName(props) {
-    const baseClass = "mdc-grid-list";
-    const classes = [];
-
-    if (
-      props[this._propsDict.padding] &&
-      props[this._propsDict.padding] === 1
-    ) {
-      classes.push("mdc-grid-list--tile-gutter-1");
-    }
-
-    if (props[this._propsDict.secondaryActionIconAlign]) {
-      classes.push(
-        "mdc-grid-list--with-icon-align-" +
-          props[this._propsDict.secondaryActionIconAlign]
-      );
-    }
-
-    if (props[this._propsDict.aspectRatio]) {
-      classes.push(
-        "mdc-grid-list--tile-aspect-" + props[this._propsDict.aspectRatio]
-      );
-    }
-
-    if (props[this._propsDict.twoLineCaptions]) {
-      classes.push("mdc-grid-list--twoline-caption");
-    }
-
-    return classes.join(" ");
+    this._mdcProps = [
+      "header-caption",
+      "twoline-caption",
+      "tile-gutter",
+      "icon-align",
+      "tile-aspect-ratio"
+    ];
   }
   materialDom(props) {
-    const cols = props.cols || 2;
-    const tileWidthPct = Math.floor(100 / cols) + "%";
+    // TODO: validate that the only values are "start" or "end"
+    let className = props["icon-align"]
+      ? `mdc-grid-list--with-icon-align-${props["icon-align"]}`
+      : "";
+
+    // TODO: validate that the only values are "1x1", "16x9", "2x3", "3x2", "4x3", or "3x4"
+    if (props["tile-aspect-ratio"]) {
+      className += ` mdc-grid-list--tile-aspect-${props["tile-aspect-ratio"]}`;
+    }
+
+    // TODO: validate that the only value is "1"
+    if (props["tile-gutter"]) {
+      className += ` mdc-grid-list--tile-gutter-${props["tile-gutter"]}`;
+    }
 
     return (
-      <span>
-        <style
-          dangerouslySetInnerHTML={{
-            __html: `
-        .mdc-grid-list__tiles {
-          --mdc-grid-list-tile-width: ${tileWidthPct};
-        }
-        .mdc-grid-list__tiles > .mdc-grid-tile > * img {
-          width: 100%;
-        }
-      `
-          }}
-        />
-        <div
-          className={this.createClassName(props)}
-          ref={control => {
-            this.control = control;
-          }}
-          {...props}
-        >
-          <ul class="mdc-grid-list__tiles">{props.children}</ul>
-        </div>
-      </span>
+      <div {...props} className={className}>
+        {props.children}
+      </div>
     );
   }
 }
 
-/**
- * @prop actionIcon	{Node} - An IconButton element to be used as secondary action target (primary action target is the tile itself).
- * @prop actionPosition {'left'|'right'} - Position of secondary action IconButton.
- * @prop children	{Node} - Theoretically you can pass any node as children, but the main use case is to pass an img, in whichcase GridTile takes care of making the image "cover" available space (similar to background-size: cover or to object-fit:cover).
- * @prop cols	{number} - Width of the tile in number of grid cells.
- * @prop containerElement	{string|node} - Either a string used as tag name for the tile root element, or a ReactElement. This is useful when you have, for example, a custom implementation of a navigation link (that knows about your routes) and you want to use it as the primary tile action. In case you pass a ReactElement, please ensure that it passes all props, accepts styles overrides and render it's children.
- * @prop rows	{number} - Height of the tile in number of grid cells.
- * @prop style {object} - Override the inline-styles of the root element.
- * @prop subtitle	{Node} - String or element serving as subtitle (support text).
- * @prop subtitleStyle {object} - Override the inline-styles of the subtitle element.
- * @prop title {Node} - Title to be displayed on tile.
- * @prop titleBackground {string} - 'rgba(0, 0, 0, 0.4)'	Style used for title bar background. Useful for setting custom gradients for example
- * @prop titlePosition {'top'|'bottom'} - Position of the title bar (container of title, subtitle and action icon).
- * @prop titleStyle {object} - Override the inline-styles of the title element.
- */
+class GridListTiles extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "grid-list__tiles";
+  }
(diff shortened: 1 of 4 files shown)


edhille:
One other thing I noticed: the .gitignore file prevents *.css files from being included, but since you are allowed to include individual stylesheets for components, I'm assuming the .gitignore rule is a recent addition that came after the majority of components were added.

In my fork, I extracted the grid-list styles from preact-material-components/node_modules/material-components-web/dist/material-components-web.css into the GridList directory, but perhaps there is a different way that stylesheet is automagically generated?

filoozom:
For CSS, you should have a look at [CssMigrationWebpackPlugin.js](https://github.com/prateekbh/preact-material-components/blob/master/CssMigrationWebpackPlugin.js). It copies CSS files from ```material-components-web``` to this library when you run ```npm run build```, and they shouldn't be commited.

So for testing you should just be able to run ```npm run build``` and then import the file like so:
```js
import 'preact-material-components/GridList/style.css';
```

GridList/GridList.jsx
```diff
@@ -0,0 +1,178 @@
...
+          </span>
+        ) : (
+          ""
+        )}
+      </li>
+    );
+  }
+  render() {
+    const element = super.render();
+    // remove the extra attributes used for customising this element - keep the DOM clean
```
filoozom: Good initiative, we'll actually be working on an easier solution for this as requested in #240.

filoozom:
First, thanks for the initiative!

Globally, I think you may not have fully understood [MaterialComponent.js](https://github.com/prateekbh/preact-material-components/blob/master/MaterialComponent.js). The thing is that ```MaterialComponent``` automatically supports ```material-components-web``` props with ```this._mdcProps```. For example, this:
```jsx
if (props[this._propsDict.twoLineCaptions]) {
  classes.push("mdc-grid-list--twoline-caption");
}
```

can just be replaced with:
```jsx
constructor() {
  super();
  this.componentName = "grid-list";
  this._mdcProps = ["twoline-caption"];
}
```
then you just add the props on the element:
```jsx
materialDom(props) {
  return (
    <div {...props}>
      {props.children}
    </div>
  );
}
```
and ```MaterialComponent``` will automatically take care of that prop for you (which means add the class to the element and remove the prop from the DOM).

Actually, you could just add all these items to ```_mdcProps``` and get rid of ```createClassName``` altogether:
- tile-gutter-*
- with-icon-align-*
- tile-aspect-*
- twoline-caption

Please look at this example: https://github.com/prateekbh/preact-material-components/blob/master/Textfield/Textfield.jsx#L48.

Do not hesitate to ask questions if you need help.

GridList/GridList.jsx
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    if (props[this._propsDict.aspectRatio]) {
+      classes.push(
+        "mdc-grid-list--tile-aspect-" + props[this._propsDict.aspectRatio]
+      );
+    }
+
+    if (props[this._propsDict.twoLineCaptions]) {
+      classes.push("mdc-grid-list--twoline-caption");
```
filoozom: Please check https://github.com/prateekbh/preact-material-components/pull/247#issuecomment-327836903 for more information on how to implement this correctly.

prateekbh:
LGTM, please add the same in README as well

GridList/GridList.jsx
```diff
@@ -0,0 +1,178 @@
...
+ * @prop style {object} - Override the inline styles of the root element
+ */
+class GridList extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "grid-list";
+    this._propsDict = {
+      cols: "cols",
+      padding: "padding",
+      cellHeight: "cellHeight",
```
prateekbh: where is cellHeight used?

GridList/GridList.jsx
```diff
@@ -0,0 +1,178 @@
...
+      );
+    }
+
+    if (props[this._propsDict.aspectRatio]) {
+      classes.push(
+        "mdc-grid-list--tile-aspect-" + props[this._propsDict.aspectRatio]
+      );
+    }
+
+    if (props[this._propsDict.twoLineCaptions]) {
```
prateekbh: this shoudl be in _mdcProps

GridList/GridList.jsx
```diff
@@ -0,0 +1,178 @@
...
+ * @prop titleBackground {string} - 'rgba(0, 0, 0, 0.4)'	Style used for title bar background. Useful for setting custom gradients for example
+ * @prop titlePosition {'top'|'bottom'} - Position of the title bar (container of title, subtitle and action icon).
+ * @prop titleStyle {object} - Override the inline-styles of the title element.
+ */
+class GridListTile extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "grid-list-tile";
+    this._propsDict = {
+      actionIcon: "actionIcon",
```
prateekbh: @filoozom shouldn't we implement this as a separate component?

GridList/GridList.jsx
```diff
@@ -0,0 +1,178 @@
...
+      actionPosition: "actionPosition",
+      children: "children",
+      cols: "cols",
+      containerElement: "containerElement",
+      rows: "rows",
+      subtitle: "subtitle",
+      subtitleStyle: "subtitleStyle",
+      title: "title",
```
prateekbh: all nodes should be separate components check `select`/`card` for example

GridList/GridList.jsx
```diff
@@ -0,0 +1,178 @@
...
+ * @prop cols	{number} - Width of the tile in number of grid cells.
+ * @prop containerElement	{string|node} - Either a string used as tag name for the tile root element, or a ReactElement. This is useful when you have, for example, a custom implementation of a navigation link (that knows about your routes) and you want to use it as the primary tile action. In case you pass a ReactElement, please ensure that it passes all props, accepts styles overrides and render it's children.
+ * @prop rows	{number} - Height of the tile in number of grid cells.
+ * @prop style {object} - Override the inline-styles of the root element.
+ * @prop subtitle	{Node} - String or element serving as subtitle (support text).
+ * @prop subtitleStyle {object} - Override the inline-styles of the subtitle element.
+ * @prop title {Node} - Title to be displayed on tile.
+ * @prop titleBackground {string} - 'rgba(0, 0, 0, 0.4)'	Style used for title bar background. Useful for setting custom gradients for example
+ * @prop titlePosition {'top'|'bottom'} - Position of the title bar (container of title, subtitle and action icon).
+ * @prop titleStyle {object} - Override the inline-styles of the title element.
```
prateekbh: where are these used?

GridList/GridList.jsx
filoozom: I think I'd prefer to have ```GridTile```s in a ```GridTileList``` that so that I could write something like:
```jsx
<GridTileList>
  <GridTile primary>
    <GridTile src="">
  </GridTile>
  <GridTile secondary>
    <GridTile title>Title</GridTile>
  </GridTile>
</GridTileList>
```

It would make the whole thing more modular. Haven't studied the doc yet though, so it might not make sense.

GridList/GridList.jsx
filoozom: Yes, exactly what I meant in https://github.com/prateekbh/preact-material-components/pull/247#discussion_r137579349.

GridList/GridList.jsx
```diff
@@ -0,0 +1,178 @@
...
+    return classes.join(" ");
+  }
+  materialDom(props) {
+    const cols = props.cols || 2;
+    const tileWidthPct = Math.floor(100 / cols) + "%";
+
+    return (
+      <span>
+        <style
+          dangerouslySetInnerHTML={{
```
filoozom: I don't think that this belongs in the components. Wouldn't it be far cleaner to just add this to an external CSS?

GridList/GridList.jsx
edhille: That is great. I will keep an eye on that and if it gets done before this is ready to merge, I'll definitely incorporate that change.

GridList/GridList.jsx
edhille: Not yet.  There are a lot of unused props at this point because I assumed we would want to mirror the interface from Material UI (http://www.material-ui.com/#/components/grid-list) and so I stubbed out the props they supported.

GridList/GridList.jsx
edhille: This raises an important question for me: how much do we want to align with the Material UI version (http://www.material-ui.com/#/components/grid-list)?

An example from that page is:
```
   <GridList
      cellHeight={180}
      style={styles.gridList}
    >
      <Subheader>December</Subheader>
      {tilesData.map((tile) => (
        <GridTile
          key={tile.img}
          title={tile.title}
          subtitle={<span>by <b>{tile.author}</b></span>}
          actionIcon={<IconButton><StarBorder color="white" /></IconButton>}
        >
          <img src={tile.img} />
        </GridTile>
      ))}
    </GridList>
```

From this comment and the next, it sounds like the preference is for something like:
```
<GridList>
  {tilesData.map((tile) => (
      <GridTile key={tile.img} actionIcon={<IconButton><StarBorder color="white"/></IconButton>}>
             <GridTileTitle>{tile.title}</GridTileTitle>
             <GridTileSubTitle>
                <span>by <b>{tile.author}</b></span>
             </GridTileSubTitle>
             <img src={tile.img}/>
      </GridTile>
  ))}
</GridList>
```

GridList/GridList.jsx
edhille: see reply to [#247 (comment)](https://github.com/prateekbh/preact-material-components/pull/247#discussion_r137579349)

GridList/GridList.jsx
edhille: nowhere yet. they are stubs taken from http://www.material-ui.com/#/components/grid-list under the assumption we'd want to be compatible with that api

GridList/GridList.jsx
edhille: Yes, I struggled with this one. I'm not a fan of this, but I'm not sure how we could support the `cols` prop that is supported http://www.material-ui.com/#/components/grid-list without something like this.

I do believe that it's a nicer interface to say, "give me a gridlist that has X columns" rather than having to add your own style rule, particularly if you are generating this programmatically.

However, if there is a strong desire to avoid this sort of embedded styles, I can remove it.

GridList/GridList.jsx
edhille: That is indeed more elegant, but I think I want to clarify one thing: it appears that `_mdcProps` is a way to allow for easy boolean-based classname setting?  In that case, it would a perfect fit for

```
    if (props[this._propsDict.twoLineCaptions]) {
      classes.push("mdc-grid-list--twoline-caption");
    }
```

Since that is boolean, it could be replaced with `this._mdcProps = ["twoline-caption"];` and then could be activated by defining an instance as `<GridList twoline-caption=true>`.

However, for tile-gutter-*, tile-aspect-*, and with-tile-icon-*, since the `*` is a placeholder for a value, I'm assuming we still need something like the `createClassName` method correct?

GridList/GridList.jsx
prateekbh: actually we're more focussed to be as close as possible to `mdc-web`. The project intends to be a thin declarative wrapper over it for preact.

GridList/GridList.jsx
edhille: @filoozom did I understand how `_mdcProps` works?

GridList/GridList.jsx
edhille: so am I to understand that the second example I gave is the tag structure you would prefer to use?

GridList/GridList.jsx
edhille: @filoozom / @prateekbh - what are your thoughts about embedding the style to make it more declarative?

filoozom: Well, it would work with SSR so there's that, but I'm really not a fan of it.

GridList/GridList.jsx
filoozom: Yes, we'd prefer the second one, with a ```GridTilePrimary``` for the ```img```.

edhille:
Closing and will issue a new, revised PR

PR #248, opened 2017-09-08
Only deploy for NodeJS 8 and docs environment

filoozom:
Changes three things:
- Make sure to deploy the right folder (hopefully environment variables are supported there, documentation is not quite clear on that)
- Only deploy for NodeJS 8
- Only deploy when we are actually testing docs and not the root repo


prateekbh:
lets hope so

PR #251, opened 2017-09-08
adding base tag in template

prateekbh:
adding template and base tag

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ cache:
     - node_modules
     - docs/node_modules
 
-before_deploy: cd $TRAVIS_BUILD_DIR/docs && npm run build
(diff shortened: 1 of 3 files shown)


filoozom:
Maybe add back the lines from the old ```index.html```:
```html
<meta name="viewport" content="initial-scale=1, maximum-scale=1">
<meta name="keywords" content="preact, material components, preact components">
<meta name="description" content="preact-material-components is a thin opinionless wrapper around material-components-web. Use these components to add material components to your web app, without worrying about the extra code of the the components which you are not using.">
<link rel="shortcut icon" type="image/png" href="favicon.png">
<link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```

PR #254, opened 2017-09-10
Add template

prateekbh:
Changed the domain and added new login and token in travis.

I've also deleted the GITHUB_TOKEN, we dont still need it right?


.travis.yml
```diff
@@ -20,8 +20,8 @@ before_deploy: cd $TRAVIS_BUILD_DIR/docs && npm run build:prod
 
 deploy:
   provider: surge
-  project: ./docs/build
-  domain: preact-material-components.surge.sh
+  project: $TRAVIS_BUILD_DIR/docs/build
```
filoozom: This should be reverted.

PR #256, opened 2017-09-10
fixing description

prateekbh:
To be merged after #255 
fixes a small description and checks up our auto deploy


filoozom:
@prateekbh The old way of adding ```z1={true}``` still works though. Should maybe remove it but it's a breaking change.

prateekbh:
its more like deprecating for now

PR #261, opened 2017-09-17
Revised GridList implementation

edhille:
Here is a revised version of the GridList per feedback from the initial PR.


GridList/GridList.jsx
```diff
@@ -0,0 +1,153 @@
...
+};
+
+const isValidValue = (validationValues, testValue) => {
+  return (
+    validationValues &&
+    validationValues.findIndex(val => val === testValue) >= 0
+  );
+};
+
+const notEmptyString = val => val !== "";
```
filoozom: Why are these constants / functions not in the ```GridList``` class?

GridList/GridList.jsx
```diff
@@ -0,0 +1,153 @@
...
+class GridList extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "grid-list";
+    this._mdcProps = [
+      "header-caption",
+      "twoline-caption",
+      "tile-gutter",
+      "with-icon-align",
+      "tile-aspect-ratio"
```
filoozom: This would allow for invalid values, for example just writing ```<GridList tile-aspect-ratio>``` would add ```className="tile-aspect-ratio"``` on the element. Guess it's not that huge of an issue and it's a good idea to use ```_mdcProps``` to clean the element, so that's actually pretty nice.
Ping @prateekbh to see if we shouldn't add this pattern to ```MaterialComponent```:
```js
const VALIDATION_VALUES_BY_KEY = {
    "with-icon-align": ["start", "end"],
   "tile-gutter": [1],
   "tile-aspect-ratio": ["1x1", "16x9", "2x3", "3x2", "4x3", "3x4"]
 };
```

GridList/GridList.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+class GridListTilePrimaryContent extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "grid-tile__primary-content";
+  }
+  materialDom(props) {
+    return (
+      <div>
+        <img {...props} />
```
filoozom: Here you should just:
```js
return (
    <img {...props} />
);
```
because ```grid-tile__primary-content``` should be set on the ```img``` and not the ```div```. Also, the ```div``` is completely unnecessary.

GridList/GridList.jsx
edhille: No practical reason. Purely code style. They can be moved in if preferred.

GridList/GridList.jsx
edhille: Good catch. Thanks!

GridList/GridList.jsx
edhille: I moved them into the class. Is this what you were envisioning?

GridList/GridList.jsx
```diff
@@ -0,0 +1,151 @@
...
+    );
+  }
+
+  constructor() {
+    super();
+    this.componentName = "grid-list";
+    this._mdcProps = [
+      "header-caption",
+      "twoline-caption",
+      "tile-gutter",
```
prateekbh: can we keep this `tile-gutter-1`? that ways it becomes super easy and less code.

GridList/GridList.jsx
```diff
@@ -0,0 +1,151 @@
...
+      "with-icon-align",
```
prateekbh: if we want this to be a multi value prop, lets remove it from here.
`mdc-props` is the home to boolean props for easy plug n play config

GridList/GridList.jsx
prateekbh: this shouldn't really be here  since its a multi value prop.

also about clean up of props, i am starting to work on clean DOM PR soon, lets wait for it.

GridList/GridList.jsx
```diff
@@ -0,0 +1,151 @@
...
+class GridListTileIcon extends Icon {
+  constructor() {
+    super();
+    this.componentName = "grid-tile__icon";
+  }
+}
+
+GridList.Tiles = GridListTiles;
+GridList.Tile = GridListTile;
+GridList.Tile.Primary = GridListTilePrimary;
```
prateekbh: this might make the DOM a lil ugly, can we restrict to single level sub-nesting only?
`Grid.PrimaryTile`

GridList/GridList.jsx
```diff
@@ -0,0 +1,151 @@
...
+GridList.Tile.Primary.Content = GridListTilePrimaryContent;
```
prateekbh: suggestion: `Grid.PrimaryContentTile`

GridList/GridList.jsx
```diff
@@ -0,0 +1,151 @@
...
+GridList.Tile.Secondary = GridListTileSecondary;
```
prateekbh: suggestion: `Grid.SecondaryTile` and so on for other components too

GridList/GridList.jsx
filoozom: Yes but ```_mdcProps``` is still useful to remove the props from the DOM.

GridList/GridList.jsx
edhille: I guess. Since it is a number, I assumed this was something that could change in the future and at that point it would be a breaking change.

GridList/GridList.jsx
edhille: To me this loses the implied structural hierarchy in the names (a GridTile and a GridIconTile sound like different flavors of the same thing rather than the GridIconTile needing to be a child of the GridTile), but I've renamed them per your suggestion.

prateekbh:
@edhille can you add a change to bump the version of package.json?
also will you be taking up the documentation or do you want to open an issue with `help-wanted` label?

edhille:
Bumped the version and added documentation, but I need help with one thing: I could not get the image I added for use in the demo to actually get included. If I could get help with that, it would be great.

filoozom:
Yeah, we need a more robust alternative to ```copyAssets``` in ```package.json```. For now, you can just replace it with:
```
"copyAssets": "cp -R images *.png *.json build/",
```
I guess.

GridList/GridList.jsx
filoozom: Yes, thanks.

edhille:
hmm...I must be missing something really simple: I changed the command, the file was copied, but it's still not showing up. Looks like we are using standard `preact` which defaults to serving everything under `/build`, so I'm not sure what could be going on here.

prateekbh:
@edhille thanks for the awesome contribution. Your changes are published and docs are updated 🥂 🥂 🥂.

Issue #263, opened 2017-09-19
Migrate to mdc 0.21.0


prateekbh:
This will have a longer lists of tasks as many things have changed

- [x] 1. Button: Remove primary/accented props
- [x] 2. Button: Implement `mdc-button__icon`
- [ ] 3. Update Button's documentation
- [x] 4. FAB: implement `mdc-fab--exited`
- [x] 5. FAB: remove `mdc-fab--plain`
- [x] 6. FAB: implement `fab__icon`
- [ ] 7. Update Fab's documentation 
- [x] 8. Snackbar: Removed the dependency of mdc-button from DOM structure of snackbar
- [x] 9. Update Snackbar's documentation
- [x] 10. Add `.scss` files in the package as there is a lot of dependency here.
- [ ] 11. Add a page for themeing, as is becomes really important to tell users how to play with colors since primary/accented are gone

P.S. lets first merge all of this in a branch `migrate21` and then merge it to master

prateekbh:
CC: @filoozom in case you are prepping to pick any of these

prateekbh:
#266  fixes 1 and 2

prateekbh:
#267 fixes 4, 5, 6

prateekbh:
#268 fixes number 10

prateekbh:
#272 fixes number 8 and no documentation update is needed for this

PR #266, opened 2017-09-19
fixing button for mdc0.21.0

prateekbh:
fixes 1 and 2 from #263

### DIFF
--- a/Button/Button.jsx
+++ b/Button/Button.jsx
@@ -1,6 +1,7 @@
 import { h } from "preact";
 import MaterialComponent from "../MaterialComponent";
 import Icon from "../Icon/";
+
 /**
  *  @prop dense = false
  *  @prop raised = false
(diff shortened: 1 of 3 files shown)


Button/Button.jsx
```diff
@@ -1,28 +1,19 @@
 import { h } from "preact";
 import MaterialComponent from "../MaterialComponent";
-
+import Icon from "../Icon/";
 /**
```
filoozom: Tiny detail but maybe add a new line after this 😛

prateekbh:
@filoozom done!

prateekbh:
@filoozom if i updated docs now and merged it to master, it'll search for the new version before its published and might break... isnt it?

filoozom:
Oh, right, didn't even think about that... We might need a deploy script for ```npm``` or push the commit after the release, but that actually doesn't make it any easier to update docs at the same time 😞.

prateekbh:
@filoozom if we do that, the problem will be that there might be a point where it does not point to the published version but the the one in WIP. this might confuse people.

PR #267, opened 2017-09-19
fixes fab for mdc0.21.0

prateekbh:
fixes 4 5 6 from #263


prateekbh:
@filoozom can you LG this one as well?

PR #277, opened 2017-09-27
temporary fix for minified code example

prateekbh:
temporary fix for minified code in docs


filoozom:
You could do something like:
```
const jsxToString = jsx => require('jsxToString')(jsx, {
    displayName: 'Button'
})
```
I think.

prateekbh:
not sure though how, a complex nested elements like <Select> will work with this

filoozom:
Right. Strange that it worked on your local version.

prateekbh:
nah i was using `preact-watch`. broke for me too with `preact serve`

PR #282, opened 2017-09-29
Demo of the Drawer

keemor:
Demo of the Drawer

### DIFF
--- a/docs/src/routes/drawer/index.js
+++ b/docs/src/routes/drawer/index.js
@@ -3,6 +3,8 @@ import { h, Component } from "preact";
 
 // Material Components
 import Drawer from "preact-material-components/Drawer";
+import Button from "preact-material-components/Button";
+import List from "preact-material-components/List";
 
 // Components
 import ComponentTable from "../../components/component-table";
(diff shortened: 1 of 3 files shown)


docs/src/routes/drawer/index.js
```diff
@@ -83,7 +85,32 @@ export default class DrawerPage extends Component {
         </div>
 
         <div className="mdc-typography--display1">Demo </div>
-        <div className="mdc-typography--title">not available </div>
+        <Button
```
prateekbh: can you try to contain the drawer into a smaller container div?

keemor: What should be the effect of this change?
Now the result is like on this example: http://www.material-ui.com/#/components/drawer
and I followed this example https://material.preactjs.com/component/dialog

prateekbh: i am not able to find a live example, but consider something like a drawer opening/closing inside a div viewport

keemor: I added drawer toggling inside div

PR #294, opened 2017-10-13
making menu declarative

prateekbh:
this adds a declarative `open` prop to menu and adds events as props too

### DIFF
--- a/Menu/Menu.jsx
+++ b/Menu/Menu.jsx
@@ -3,6 +3,13 @@ import MaterialComponent from "../MaterialComponent";
 import { MDCSimpleMenu } from "@material/menu";
 import List from "../List";
 
+/*
+ * Default props for menu
+ */
+const defaultProps = {
+  open: false
+};
+
 /**
  * @prop open = false
  */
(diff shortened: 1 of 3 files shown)


Menu/Menu.jsx
```diff
@@ -17,12 +24,39 @@ class Menu extends MaterialComponent {
...
+    this._cancel = this._cancel.bind(this);
   }
   componentDidMount() {
     this.MDComponent = new MDCSimpleMenu(this.control);
+    this.MDComponent.listen("MDCSimpleMenu:selected", this._select);
+    this.MDComponent.listen("MDCSimpleMenu:cancel", this._cancel);
+    toggleMenu(defaultProps, this.props, this.MDComponent);
   }
   componentWillUnmount() {
     this.MDComponent.destroy && this.MDComponent.destroy();
```
filoozom: ```destroy``` should probably be called after ```unlisten```. Also I'm not sure if ```unlisten``` is necessary, I'll have a look at the code.

prateekbh:
I kinda figured out that while we are a wrapper for preact, its good to expose the underlying API in the declarative way `preact/react` devs are used to. moreover its still a wrapper over underlying API. 
still custom code is kept to min

Menu/Menu.jsx
prateekbh: fixed it, although unlisten is still there, for cleanup sake

PR #295, opened 2017-10-15
Support custom `class` prop

developit:
Example: `<Card.Title class="foo" />`
Currently only `className` is supported.


filoozom:
Just saw that we don't run ```prettier``` on ```MaterialComponent.js```, but I'll open a separate PR for that.

PR #296, opened 2017-10-16
fixing class prop messup

prateekbh:
fixes  a small bug to support class as a prop

### DIFF
--- a/MaterialComponent.js
+++ b/MaterialComponent.js
@@ -60,7 +60,12 @@ export default class MaterialComponent extends Component {
   render() {
     this.buildClassName();
     // Fetch a VNode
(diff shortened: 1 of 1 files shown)


MaterialComponent.js
```diff
@@ -60,7 +60,12 @@ export default class MaterialComponent extends Component {
   render() {
     this.buildClassName();
     // Fetch a VNode
-    const element = this.materialDom(this.props);
+    const componentProps = this.props;
+    if (componentProps.class){
```
filoozom: Is this condition really necessary? Might as well just call delete every time I think.

prateekbh: may be.. just a cautious check TBH

Issue #315, opened 2017-10-17
Snackbar conflicts with update of prop/state


eligao:
Hi all,

I have encountered an issue when using Snackbar with dynamic contents.
When snackbar is shown, any update of prop/state will render a broken snackbar.
Sometimes the snackbar never disappears, sometimes shows no text.

I have created a minimal sandbox reproducing this problem 
[HERE](https://codesandbox.io/s/7w20pml681)
Simply click the button, you can see a broken snackbar.
But if you remove the `setState()`, the snackbar renders OK.

I assume it might be triggered by unnecessary rerendering on Snackbars.
Maybe we could somehow prevent the snackbars from re-rendering with `shouldComponentUpdate()`?

eligao:
Hi, I've opened a PR at https://github.com/prateekbh/preact-material-components/pull/317

PR #317, opened 2017-10-20
Fix incorrect Snackbar render caused by updates

eligao:


eligao:
This fixes #315 by making Snackbar a pure component, checking if states/props are changed before updating.
I'm not sure whether I should move util the function like `shallowDiffers()` to a separate file, but other components might also have similar problem.
Also, this can be done by importing `preact-compat/PureComponent`, but I'm not sure if it worth it importing a dependency only for a fraction of code.

eligao:
BTW I somehow couldn't run the pre-commit hook, it always says `./node_modules/pre-commit/hook: line 49: : command not found`, so I manually ran lint and copied the linted file to submit on Github web.
Also the version of `preact-material-components` in `package-lock.json` seem to be outdated.

Snackbar/Snackbar.jsx
```diff
@@ -1,13 +1,21 @@
...
+}
+
 /**
  * dismissesOnAction = true
  */
 export default class Snackbar extends MaterialComponent {
   constructor() {
     super();
     this.componentName = "snackbar";
+    this.isPureReactComponent = true;
```
prateekbh: this isn't required right?

Snackbar/Snackbar.jsx
```diff
@@ -23,6 +31,11 @@ export default class Snackbar extends MaterialComponent {
   componentWillUnmount() {
     this.MDComponent.destroy && this.MDComponent.destroy();
   }
+  shouldComponentUpdate(props, state) {
+    return (
+      shallowDiffers(this.props, props) || shallowDiffers(this.state, state)
```
prateekbh: doesn't preact do this on its own?

Snackbar/Snackbar.jsx
eligao: Probably not, I copied from `preact-compat/PureComponent` class

Snackbar/Snackbar.jsx
eligao: Not in `preact`, but in `preact-compat`.
That's why I said importing and inheriting from `PureComponent` from `preact-compat` may also solve this problem. However I'm not sure if it worth it importing hundreds of lines of codes when using only less than 10 lines of them.
Probably webpack tree shaking can help though.

reference:
https://github.com/developit/preact/blob/0a88752e4890d391b5e23809d9432f625a833ae8/src/vdom/component.js#L83

https://github.com/developit/preact-compat/blob/5ca57489b75c539a0ad022f3624d97d755c939b5/src/index.js#L591

PR #323, opened 2017-10-24
fixing menu icon

prateekbh:
migrates to `mdc-web` 0.23.0

### DIFF
--- a/Toolbar/Toolbar.jsx
+++ b/Toolbar/Toolbar.jsx
@@ -79,7 +79,7 @@ class ToolbarIcon extends MaterialComponent {
     super();
     this.componentName = "toolbar__icon";
     if (props.menu) {
-      this.componentName += "-menu";
+      this.componentName += "-menu-icon";
     }
   }
   materialDom(props) {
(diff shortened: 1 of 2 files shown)


Toolbar/Toolbar.jsx
```diff
@@ -79,7 +79,7 @@ class ToolbarIcon extends MaterialComponent {
     super();
     this.componentName = "toolbar__icon";
     if (props.menu) {
-      this.componentName += "--menu";
+      this.componentName += "-menu";
```
filoozom: The final output should be ```mdc-toolbar__menu-icon```, not ```mdc-toolbar_icon-menu```.

https://github.com/material-components/material-components-web/blob/master/CHANGELOG.md#breaking-changes

filoozom: Maybe a new component would be cleaner than completely changing ```componentName```?

prateekbh:
Thanks for catching that 👍

filoozom:
Still not right though 😛 
Should be ```this.componentName = "toolbar__menu-icon";```. Now it's ```mdc-toolbar__icon-menu-icon```.

prateekbh:
@filoozom what would i do without ur reviews 😛 . Thanks for taking care of this

PR #327, opened 2017-10-24
Updating Docs

prateekbh:


docs/package.json
```diff
@@ -38,17 +38,19 @@
...
-    "preact-material-components": "^1.2.1",
+    "preact-cli": "^2.0.2",
+    "preact-material-components": "^1.2.3",
     "prettier": "^1.6.1",
     "sass-loader": "^6.0.6"
   },
   "dependencies": {
-    "@webpack-blocks/webpack2": "^0.4.0"
+    "preact": "^8.2.1",
+    "preact-compat": "^3.17.0",
```
filoozom: Why do we need this?

docs/package.json
```diff
@@ -38,17 +38,19 @@
     "classnames": "^2.2.5",
     "eslint": "^4.6.1",
     "eslint-config-prettier": "^2.4.0",
+    "eslint-config-synacor": "^1.1.0",
```
filoozom: Why do we need this?

docs/package.json
prateekbh: was just testing, docs started throwing errors lately

PR #329, opened 2017-10-24
Added discrete option on Slider (#324)

### LINKED ISSUE #324: Slider discrete
Would it be possible to add the discrete option?

https://github.com/material-components/material-components-web/tree/master/packages/mdc-slider#discrete-slider

porfirioribeiro:
Implemented Slider discrete option fixes #324

### DIFF
--- a/Slider/Slider.jsx
+++ b/Slider/Slider.jsx
@@ -66,7 +66,7 @@ export default class Slider extends MaterialComponent {
           <div class="mdc-slider__track" />
         </div>
         <div class="mdc-slider__thumb-container">
-          {allprops.discrete && (
+          {props.discrete && (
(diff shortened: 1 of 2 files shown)


Slider/Slider.jsx
```diff
@@ -66,6 +66,11 @@ export default class Slider extends MaterialComponent {
           <div class="mdc-slider__track" />
         </div>
         <div class="mdc-slider__thumb-container">
+          {allprops.discrete && (
```
prateekbh: just read this from `props` instead of `allprops`

porfirioribeiro: I changed that to props, but now i am thinking, should not be destructed allprops like tabindex?

PR #333, opened 2017-10-27
Add ability to pass "valid" as a prop to the Textfield component

### LINKED ISSUE #332: Textfield does not implement `valid` prop
I'm trying to pass a prop `valid` to `<Textfield />`, as described [here](https://material.io/components/web/catalog/input-controls/text-fields/#mdctextfield-api)

Doesn't seem to do anything. 

I would expect it to add the appropriate classes, for example `.mdc-textfield--invalid` to the top level div.

mellogarrett:
Fixes #332

### DIFF
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -91,9 +91,9 @@ class TextfieldInput extends MaterialComponent {
           />
         )}
         {props.label &&
-          this.state.showFloatingLabel && (
-            <Label for={props.id}>{props.label}</Label>
-          )}
+        this.state.showFloatingLabel && (
+          <Label for={props.id}>{props.label}</Label>
+        )}
         {props.multiline ? "" : <div class="mdc-textfield__bottom-line" />}
       </div>
     );
@@ -157,9 +157,9 @@ class Textfield extends Component {
     return showDiv ? (
       <div className={className}>
         {props.label &&
-          !showFloatingLabel && (
-            <label for={props.id}>{props.cssLabel || `${props.label}: `}</label>
-          )}
+        !showFloatingLabel && (
+          <label for={props.id}>{props.cssLabel || `${props.label}: `}</label>
+        )}
         <TextfieldInput
           {...props}
           onInit={MDComponent => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7 @@
 {
   "name": "preact-material-components",
-  "version": "1.2.1",
+  "version": "1.2.4",
   "lockfileVersion": 1,
-  "requires": true,
   "dependencies": {
     "@material/animation": {
       "version": "0.4.0",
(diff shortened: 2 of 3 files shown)


Textfield/Textfield.jsx
```diff
@@ -115,11 +119,26 @@ class Textfield extends Component {
   }
 
   componentDidMount() {
+    this._setInvalid(defaultProps, this.props, this.MDComponent);
     this.setState({
       showFloatingLabel: true
     });
   }
 
+  _setInvalid(oldprops, newprops, textfield) {
```
prateekbh: this does not need to be a part of the class

mellogarrett:
Hmm. `textfield` is `undefined` when `setInvalid` is called from `componentDidMount`...

mellogarrett:
Okay check it out now. It's been tested and is working for both starting with `valid: false` and also setting it to `false` after initial mount.

Textfield/Textfield.jsx
```diff
@@ -119,24 +120,13 @@ class Textfield extends Component {
...
-      "valid" in newprops &&
-      oldprops.valid !== newprops.valid
-    ) {
-      textfield.valid = newprops.valid;
-    }
-  }
-
   componentWillUpdate(nextProps) {
-    this._setInvalid(this.props, nextProps, this.MDComponent);
+    setValid(this.props, nextProps, this.MDComponent);
```
prateekbh: This class doesn't have any `MDComponent`, `TextfieldInput` has. Shouldn't this be there?

mellogarrett: Yes, that is correct

prateekbh:
wow prettier is failing! i'll have a look

Textfield/Textfield.jsx
```diff
@@ -83,9 +91,9 @@ class TextfieldInput extends MaterialComponent {
           />
         )}
         {props.label &&
-        this.state.showFloatingLabel && (
-          <Label for={props.id}>{props.label}</Label>
-        )}
+          this.state.showFloatingLabel && (
+            <Label for={props.id}>{props.label}</Label>
+          )}
```
filoozom: I assume this is breaking ```prettier```.
Can you run ```npm run prettier:write``` and push?

mellogarrett: `npm run prettier:write` did not result in any committable changes

filoozom: Why did this change anyway? Was it ```prettier``` or did you change this by hand?

mellogarrett: Prettier is changing this. Should I turn off Prettier and push the old version?

filoozom: For debugging purposes:
- what ```prettier``` version do you have?
- what OS are you on?

Don't know yet why it wants to change it on your computer but not on Travis.

mellogarrett: I am using the Prettier VS Code plugin, v: 0.24.0 on MacOS Sierra (10.12.6)

filoozom: Haven't tried the plugin yet, maybe that's the issue?

filoozom: We actually don't have a ```prettier``` config file. The VS Code  plugin is looking for those :
- package.json - prettier key
- .prettierrc
- .prettier.config.js

mellogarrett: Hmm. How do we move forward?

filoozom: I guess you could still try disabling the plugin or running ```npm run prettier:write``` with VS Code closed, and if that works we might need to add a config file.

mellogarrett: I ran `npm i` to install Prettier locally in the project. Saved it and it fixed the formatting discrepancy.

PR #336, opened 2017-10-30
Added example code for different options to Checkbox component

mellogarrett:
Here's a preview of what the changes look like:
screen shot 2017-10-30 at 9 20 45 am

### DIFF
--- a/docs/src/routes/checkbox/index.js
+++ b/docs/src/routes/checkbox/index.js
@@ -15,6 +15,9 @@ import "./style";
 
 // Samples
 import sample from "./sample.txt";
+import basicExample from "./basic-example.txt";
+import indeterminateExample from "./indeterminate-example.txt";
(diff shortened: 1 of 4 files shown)


docs/src/routes/checkbox/index.js
```diff
@@ -62,39 +62,74 @@ export default class CheckboxPage extends Component {
...
-          onClick={() => {
-            this.cb.MDComponent.indeterminate = true;
-          }}
-        >
-          Make indeterminate
-        </Button>
+        <CodeBlock>
+          <code class="lang-html">
+            {`
+<Formfield>
```
prateekbh: can we extract these into a separate file? cuz `<pre>` tag is white space sensitive

mellogarrett: Sure, no problem

PR #406, opened 2017-11-17
Fixed Textfield after mdc upgrade

porfirioribeiro:
After MDC upgrade textfield was renamed to text-field, maybe we should also change Textfield to TextField @prateekbh?

### DIFF
--- a/Textfield/Textfield.jsx
+++ b/Textfield/Textfield.jsx
@@ -1,6 +1,6 @@
 import { h, Component } from "preact";
 import MaterialComponent from "../MaterialComponent";
-import { MDCTextfield } from "@material/textfield/";
+import { MDCTextField } from "@material/textfield";
 
 /**
  * @prop persistent = false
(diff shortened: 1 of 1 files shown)


prateekbh:
Yep we should also change the name of the text field

filoozom:
Actually the folder / file should have been renamed to `TextField` too. I'll change that in the `FormField` PR.

filoozom:
`componentsList.js` is missing the change too for `CssMigrationWebpackPlugin`.

prateekbh:
@filoozom actually i am sending a PR for other changes of 0.25 I guess I can pick this up

prateekbh:
@filoozom moreover the package name hasn't changed so i dont think its needed in `componentsList.js`

filoozom:
@prateekbh I already changed `Textfield` to `TextField` in a PR comming in a few minutes. I'll have a look for the package name.

PR #409, opened 2017-11-20
upgrade to v25

prateekbh:
other changes from v25


docs/src/components/menu/index.js
```diff
@@ -62,7 +63,7 @@ export default class Menu extends Component {
     return (
       <Drawer.TemporaryDrawer ref={drawer => (this.drawer = drawer)}>
         <Drawer.TemporaryDrawerHeader className="mdc-theme--primary-bg">
-          Components
+          Components @ (v{json.devDependencies["preact-material-components"]})
```
filoozom: Nice one! Shouldn't `json-loader` be configured as a module in `webpack` though?

docs/package.json
```diff
@@ -41,9 +41,10 @@
     "eslint-plugin-prettier": "^2.2.0",
     "highlight.js": "^9.12.0",
     "if-env": "^1.0.0",
+    "json-loader": "^0.5.7",
```
filoozom: Actually, I think that we don't need this:

> Since webpack >= v2.0.0, importing of JSON files will work by default. You might still want to use this if you use a custom file extension. See the v1.0.0 -> v2.0.0 Migration Guide for more information

prateekbh: yeah read this later...

Icon/Icon.jsx
```diff
@@ -10,8 +10,12 @@ export default class Icon extends MaterialComponent {
     this.componentName = "icon";
   }
   materialDom(props) {
+    const classes = "material-icons";
```
filoozom: Should be
```jsx
const classes = ["material-icons"];
```

prateekbh: Done

PR #410, opened 2017-11-20
Rename Formfield to FormField and fix TextField rename

filoozom:


filoozom:
I see, `git` doesn't rename folders if only the case changed. Fixed.

PR #412, opened 2017-11-21
TextField updates

mellogarrett:
- Corrected prop `multiline` to `textarea` to match mdc classnames.
- Added error message and deleting label value when trying to use with `fullwidth` text fields. 
    ^ this logs the message 3 times in the sample project. maybe this logic shouldn't be in the `materialDom` function?


TextField/TextField.jsx
```diff
@@ -78,10 +78,16 @@ class TextFieldInput extends MaterialComponent {
     if ("value" in props && this.state.showFloatingLabel) {
       className = [className, "mdc-text-field--upgraded"].join(" ");
     }
+    if (props.label && props.fullwidth) {
```
prateekbh: not really sure about this

a. This will error even in production version

b. this might confuse devs as in what deleted the label

may be we can just leave a note in docs

mellogarrett: Would you rather delete the label and _not_ log the warning?

filoozom: To be honest, I'd rather log an error in every case, doesn't hurt production in my opinion but makes it a lot better for developer experience. Wouldn't remove the label though, so you'd at least see that there's a problem.

My first instinct would be the check the console to see if there are no errors, but it might not be the case for everyone. Might even throw an exception so that debuggers pick it up?

filoozom:
> Added error message and deleting label value when trying to use with fullwidth text fields.
^ this logs the message 3 times in the sample project. maybe this logic shouldn't be in the materialDom function?

Not that big of an issue I'd say, it's just there for debugging purposes after all. But yeah, `materialDom` is rendered everytime props change, which might be a few times until the first render is complete. Might be possible to optimize performance though, haven't quite looked into that.

TextField/TextField.jsx
mellogarrett: Throwing an exception stops execution. Just leaves them at an empty screen

filoozom: Well, I know, but at least it doesn't fail "silently" and the user immediatly knows that something's wrong. That's the way Angular does it. Granted, they make the error quite a bit more beautiful but still.

I don't know if it's an ideal solution though, just the one that seems easiest to debug I guess?

prateekbh: actually failing silently isn't a big issue as `mdc-web` doesn't prohibit this behaviour. Lets just console this and leave for the user if he still wants it and want to style it in his own way

prateekbh:
@mellogarrett @filoozom I'd like to remove any sort of errors and just do the console log and leave it as is

mellogarrett:
@prateekbh done. Let me know if that's what you had in mind

Issue #280, opened 2017-09-27
Add TypeScript declaration file


YePpHa:
This will make it possible to use this library in TypeScript projects.

prateekbh:
i'll try to look at this, but i have like 0 experience in this so might take long

YePpHa:
Okay, I might consider making a PR then.

prateekbh:
awesome and welcome 👍

cvuorinen:
Any progress on this yet? I'm interested in using this project with TypeScript as well, and I could help with the typings.

prateekbh:
nothing started as of now, feel free to begin a PR.
@YePpHa have u started any work already?

YePpHa:
Yes, I've started some work on this, but it's not completed yet. I've written about 50% of declaration files for the components. However, one thing that I haven't started on yet is creating a declaration for the `MDComponent` property that some of the components have.

Progress is here: https://github.com/YePpHa/preact-material-components/tree/typescript-declaration

cvuorinen:
Great, looking good. I'll try them out later when I have time, but one question. Is there a reason for declaring inherited methods like `componentDidMount` and `materialDom` on each component? Seems to me like those should be redundant since they are already inherited.

YePpHa:
You're right. Private methods are just redundant.

The reason I added the `componentDidMount` and `materialDom` is because they override the method from the class they extend from and therefore they're not the method that was declared in `MaterialComponent.d.ts`. Though, I'm not sure if it actually does any difference, but I've added them for technicalities.

cvuorinen:
If the overriding method does not change the method signature I think they are redundant from the type declaration point of view. The declared type will have the same method either way.

porfirioribeiro:
I'm also using this library with typescript, i will try to help with this also

PR #416, opened 2017-11-24
Implement TypeScript declaration files

YePpHa:
Resolves: #280

I've been forced to add a few declaration files for some of the [MDC](https://github.com/material-components/material-components-web) components because some of the preact components doesn't proxy the functionality of the MDC components.


Textfield/Textfield.d.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface ITextfieldProps extends JSX.HTMLAttributes {
+  helptext?: string;
+  helptextPersistent?: string;
+  helptextValidationMsg?: string;
+  cssLabel?: string;
+}
+export interface ITextfieldState {
+  showFloatingLabel: boolean;
+}
+export default class Textfield extends MaterialComponent<ITextfieldProps, ITextfieldState> {
```
porfirioribeiro: This was renamed to TextField in #410

Slider/Slider.d.ts
```diff
@@ -0,0 +1,40 @@
+import MaterialComponent from '../MaterialComponent';
+import { VNode } from 'preact';
+import { MDCFoundation, MDCComponent } from '../MaterialComponentsWeb';
+
+export interface ISliderProps extends JSX.HTMLAttributes {
+  disabled?: boolean;
+  min?: number;
+  max?: number;
+  step?: number;
```
porfirioribeiro: Missing 
```typescript
discrete?: boolean;
```
And maybe:
```typescript
value?: number
onInput?: (value: number) => void
onChange?: (value: number) => void
```

Select/Select.d.ts
```diff
@@ -0,0 +1,32 @@
+import MaterialComponent from '../MaterialComponent';
+import { VNode } from 'preact';
+import { MDCFoundation, MDCComponent } from '../MaterialComponentsWeb';
+import { Item as ListItem } from '../List/List';
+
+export interface ISelectProps extends JSX.HTMLAttributes {
+  disabled?: boolean;
```
porfirioribeiro: Missing
`basic`
`onChange`
`selectedIndex`

prateekbh:
wow Thanks @porfirioribeiro

Select/Select.d.ts
YePpHa: Also fixed but apparently Github didn't register it.

cvuorinen:
How are these supposed to be used? Currently e.g. [Toolbar.d.ts](https://github.com/YePpHa/preact-material-components/blob/typescript-declaration/Toolbar/Toolbar.d.ts) exports `Toolbar`, `Row`, `Section`, `Icon` & `Title` separately.

So you have to import all of them and cannot just import `Toolbar` and write like the js version?

Instead of:
```jsx
import Toolbar from 'preact-material-components/Toolbar';

<Toolbar className="toolbar">
    <Toolbar.Row>
    <Toolbar.Section align-start={true}>
        <Toolbar.Icon menu={true}>menu</Toolbar.Icon>
        <Toolbar.Title>
        ...
```

Need to write?:
```jsx
import { Toolbar, Row, Section, Icon, Title } from 'preact-material-components/Toolbar';

<Toolbar className="toolbar">
    <Row>
    <Section align-start={true}>
        <Icon menu={true}>menu</Icon>
        <Title>
        ...
```

And when you want to use `Toolbar` and `Card` in the same file, you cannot just import `Title` from both of them, so would have to alias them.

Or am I missing something?

cvuorinen:
Maybe consider using namespaces similarly as e.g. preact-mdl does? https://github.com/developit/preact-mdl/blob/2.0.0/src/index.d.ts#L38:L52

cvuorinen:
I created a test repo to try these out: https://github.com/cvuorinen/preact-material-components-typings-test/pull/1

Could not get `Toolbar.Title`, `Card.Title` etc. to work (webpack complains "export was not found"), can you check it out? Also added few other comments there.

cvuorinen:
And instead of using namespaces like `preact-mdl`, I think we could also just use static properties, e.g.
```tsx
export default class Card extends MaterialComponent<ICardProps, {}> {
  static Title: typeof Title;
  ...
}
```

Elevation/Elevation.d.ts
```diff
@@ -0,0 +1,33 @@
+import MaterialComponent from '../MaterialComponent';
+import { VNode } from 'preact';
+
+export interface IElevationProps extends JSX.HTMLAttributes {
+  'z'?: number;
```
porfirioribeiro: this could be:
```
z: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | 20 | 21 | 22 | 24;
```

Elevation/Elevation.d.ts
```diff
@@ -0,0 +1,33 @@
...
+  'z0'?: boolean;
```
porfirioribeiro: @prateekbh is it really helpful to have this props `z0` over having just z={0}?

YePpHa:
Thanks for the reviews. I will go through them and make the necessary changes when I get home today.

Elevation/Elevation.d.ts
prateekbh: we can omit this in TS it's there just to avoid breaking change

prateekbh:
@cvuorinen @porfirioribeiro let me know if this has your guy's LG

YePpHa:
I've made the changes requested.

However, I get an error in `/Drawer/index.d.ts`. It says `Type `LinkItem` is not generic. It seems to be related to `typeof` not including the generic types.
```TypeScript
import List from '../List/List';
...
declare class DrawerItem extends List.LinkItem<IDrawerItemProps & JSX.HTMLAttributes> {}
```

I'm not 100% sure on how we should fix this issue. Perhaps @cvuorinen has an idea on how to properly proceed.

cvuorinen:
Overall great progress, looks much better now :+1: 
I'll try it out with my test repo.

Drawer/index.d.ts
```diff
@@ -0,0 +1,63 @@
+import MaterialComponent from '../MaterialComponent';
+import { VNode } from 'preact';
+import List from '../List/List';
+import { MDCFoundation, MDCComponent } from '../MaterialComponentsWeb';
+
+export default interface Drawer {
```
cvuorinen: This doesn't seem to work (TypeScript complains `'Drawer' only refers to a type, but is being used as a value here.` when trying to use it). I think it needs to be a class with static properties as all the other ones, even though the Drawer itself does not have any functionality.

Dialog/index.d.ts
```diff
@@ -0,0 +1,46 @@
...
+
+declare interface IBodyProps {
+  scrollable?: boolean;
+}
+declare class Body extends MaterialComponent<IBodyProps & JSX.HTMLAttributes, {}> {}
+
+declare class Footer extends MaterialComponent<JSX.HTMLAttributes, {}> {}
+
+declare interface IFooterButtonProps {
+  accept?: boolean;
```
cvuorinen: This particular prop seems problematic. It conflicts with `JSX.HTMLAttributes` where it is defined as string. When using it as boolean TypeScript complains `Type '{ accept: true; }' is not assignable to type 'HTMLAttributes'. Type 'true' is not assignable to type 'string'.`
And when I changed it to string, TS complains `Type '{ accept: "foo"; }' is not assignable to type 'IFooterButtonProps'. Type '"foo"' is not assignable to type 'boolean'.`

I'm not exactly sure how to fix it. I was wondering why you changed all those interfaces extending JSX.HTMLAttributes to just `<Props & JSX.HTMLAttributes, {}>` in the generic, did you run into similar issues with it or something?

cvuorinen:
Regarding the issue with `LinkItem` not being generic, I think it's not necessary that the type `DrawerItem` extends `List.LinkItem` even though the actual js code does. The types are mostly about detecting correct props and the type is not inheriting them so I think there is no gain in trying to extend like that? We could probably just declare the `DrawerItem` type like every other one by extending `MaterialComponent`.

Another way would be to export LinkItem class as you did previously and import it directly in Drawer. I think there's no harm in exporting additional classes besides the default, as long as the default exports similarly as the js code.

Dialog/index.d.ts
YePpHa: I experience the issue multiple places so I figured I would just do the same thing overall to make them as identical as possible.

cvuorinen: Ok, need to look into this a bit more and try to figure a way to override the type from JSX.HTMLAttributes.

YePpHa:
@cvuorinen I found out why the issue occurred. It's actually because it sets `List.LinkItem` as `any` because I renamed everything to `index.d.ts` and I forgot to change the import to `import List from '../List';` from `import List from '../List/List';`

Dialog/index.d.ts
YePpHa: As the issue with `JSX.HTMLAttributes` we might need to create an almost identical interface but just without the `accept` property.

YePpHa:
@cvuorinen I've reverted it back to just extending `JSX.HTMLAttributes`. For the cases where there's conflicts I have made a duplicate of `JSX.HTMLAttributes` but just without the properties that conflict. It's not pretty, but it's the best I could come up with. If anyone have other suggestions that's better would be awesome.

Dialog/index.d.ts
```diff
@@ -0,0 +1,180 @@
...
+  cancel(shouldNotify: boolean): void;
+}
+
+declare class MDCDialog extends MDCComponent<MDCDialogFoundation> {
+  open(): boolean;
+  show(): void;
+  close(): void;
+}
+
+interface HTMLAttributes extends PreactHTMLAttributes, JSX.DOMAttributes {
```
cvuorinen: Not a fan of having to maintain these...
I guess they don't change that often but still would be nice to find a better way. Not sure that there is though.

But at least should name it something like `HTMLAttributesWithoutAccept` or add a comment about why this is needed and how it differs from the real HTMLAttributes etc.

YePpHa: That makes sense. I will add the comments and rename them.

Dialog/index.d.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+declare class MDCDialog extends MDCComponent<MDCDialogFoundation> {
+  open(): boolean;
+  show(): void;
+  close(): void;
+}
+
+/**
+ * `IFooterButtonProps` conflicts with `JSX.HTMLAttributes` as
```
porfirioribeiro: Maybe i am seeing this wrong, but couldn't we use something like:
http://ideasintosoftware.com/typescript-advanced-tricks/
`Omit<JSX.HTMLAttributes, 'accept'>`

YePpHa: That might be the best option we have. @cvuorinen, what do you think?

cvuorinen: Yes, that looks good. :+1: I was looking for something like that myself but didn't find yet.

porfirioribeiro: Cool, i'm happy that it worked.
It's more clean with Omit 👍

Drawer/index.d.ts
```diff
@@ -0,0 +1,63 @@
...
+  DrawerItem: typeof DrawerItem;
+  TemporaryDrawer: typeof TemporaryDrawer;
+  TemporaryDrawerHeader: typeof TemporaryDrawerHeader;
+  TemporaryDrawerContent: typeof TemporaryDrawerContent;
+  PermanentDrawer: typeof PermanentDrawer;
+  PermanentDrawerHeader: typeof PermanentDrawerHeader;
+  PermanentDrawerContent: typeof PermanentDrawerContent;
+  PersistentDrawer: typeof PersistentDrawer;
+  PersistentDrawerHeader: typeof PersistentDrawerHeader;
+  PersistentDrawerContent: typeof PersistentDrawerContent;
```
cvuorinen: Missing `static` on all of these.

Dialog/index.d.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+declare class Body extends MaterialComponent<IBodyProps, {}> {}
+
+declare class Footer extends MaterialComponent<JSX.HTMLAttributes, {}> {}
+
+declare interface IFooterButtonProps extends Omit<JSX.HTMLAttributes, 'accept'> {
+  accept?: boolean;
+  cancel?: boolean;
+}
+declare class FooterButton extends Button<IFooterButtonProps, {}> {}
```
cvuorinen: This is still not working because it extends Button, which also inherits `accept: string` from HTMLAttributes (`Type '{ accept: true; }' is not assignable to type 'IFooterButtonProps & IButtonProps & IMaterialComponentProps & ComponentProps<FooterButton>'.`).

Maybe should change back to the way you had before where props interfaces don't extend anything and use `& HTMLAttributes` in the class definition? That way could extend ButtonProps here without inheriting anything other than that?

YePpHa: I've moved `JSX.HTMLAttributes` to be the default `PropsType` for `Button`. This will make it possible to override it in `Dialog/index.d.ts`.

Menu/index.d.ts
```diff
@@ -0,0 +1,32 @@
+import MaterialComponent from '../MaterialComponent';
+import { VNode } from 'preact';
+import { MDCFoundation, MDCComponent } from '../MaterialComponentsWeb';
+import List from '../List/List';
```
cvuorinen: Should import just `../List`

cvuorinen:
Okay, now it compiles the template project without errors or any modifications to the code :tada: 

But I think there are still some props missing etc. I'm working on a more comprehensive test that will hopefully catch as much missing stuff as possible.

I think we could merge this in and add any missing stuff with new pull requests, or what do you guys think? Should we try to get this as comprehensive as possible in this PR before merging?

prateekbh:
Thank you so much guys for your great work! will update soon

PR #420, opened 2017-11-27
add http-server as dev dep. add npm script to serve sample app

mellogarrett:


package.json
```diff
@@ -11,7 +11,8 @@
     "lint": "eslint index.js */*.jsx",
     "lint-staged": "lint-staged",
     "prettier": "prettier -l *.js */*.jsx sample/webpack.config.js",
-    "prettier:write": "prettier --write *.js */*.jsx sample/webpack.config.js"
+    "prettier:write": "prettier --write *.js */*.jsx sample/webpack.config.js",
+    "sample": "$(npm bin)/http-server ./sample"
```
filoozom: `$(npm bin)` doesn't work on all shells or Windows, shouldn't `http-server ./sample` work without it?

mellogarrett: Nah, can't just execute a local npm package like that.

mellogarrett:
@filoozom this new npm script should work

package.json
filoozom: Wait, why not? Works perfectly for `prettier`, `lint-staged` and `eslint` in this very same file. Also, just tested specifically for `http-server` and it works without any issues.

filoozom:
Actually, it might be better to run `webpack` before `http-serve`, wouldn't it?
I feel like it would reduce confusion. Or maybe add something like `sample:build` / `sample:serve`.

mellogarrett:
Makes sense

porfirioribeiro:
This is really helpful when developing and testing changes.
Couldn't this work with webpack-dev-server?

prateekbh:
@porfirioribeiro its just a rough play space, nothing too serious. feel free to add a webpack dev server. I just didnt felt the need to do it then

Issue #370, opened 2017-11-13
Add support for Textfield with Leading/Trailing icons


mattiagalati:
Cannot figure out how to achieve the feature "Textfield - Leading/Trailing icons" as presented here: https://material-components-web.appspot.com/textfield.html

prateekbh:
woops this is currently un supported, happy to take this as a feature request.

prateekbh:
@mellogarrett will you be able to take this up?

mellogarrett:
yea, I can work on this. @mattiagalati the link is broken. Mind updating?

prateekbh:
https://material-components-web.appspot.com/text-field.html

PR #422, opened 2017-11-27
Textfield icons

garrett-thompson:
Implemented both trailing & leading icons in textfields. Use `trailingIcon` or `leadingIcon` prop, passing the name of the material icon

### DIFF
--- a/TextField/TextField.jsx
+++ b/TextField/TextField.jsx
@@ -75,19 +75,15 @@ class TextFieldInput extends MaterialComponent {
   }
   materialDom(allprops) {
     let { className, ...props } = allprops;
+    className = className || "";
 
-    props.leadingIcon
-      ? className != null
-        ? (className +=
-            " mdc-text-field--box mdc-text-field--with-leading-icon")
-        : (className = "mdc-text-field--box mdc-text-field--with-leading-icon")
-      : null;
-    props.trailingIcon
-      ? className != null
-        ? (className +=
-            " mdc-text-field--box mdc-text-field--with-trailing-icon")
-        : (className = "mdc-text-field--box mdc-text-field--with-trailing-icon")
-      : null;
+    if ("leadingIcon" in props) {
+      className += " mdc-text-field--box mdc-text-field--with-leading-icon";
+    }
+
+    if ("trailingIcon" in props) {
+      className += " mdc-text-field--box mdc-text-field--with-trailing-icon";
+    }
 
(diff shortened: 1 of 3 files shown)


mellogarrett:
Fixes #370

prateekbh (changes requested):
Can we use `<Icon>` instead?

mellogarrett:
The material-components-web docs use the `<i>` tag, but I can definitely change it if you think we should.

https://material.io/components/web/catalog/input-controls/text-fields/#leading-and-trailing-icons

prateekbh:
@mellogarrett doesnt `<Icon>` renders to `<i>`?

mellogarrett:
Oh sorry, I misunderstood. I'll make that change.

prateekbh:
last thing, can you bump the version?

package.json
```diff
@@ -90,14 +90,8 @@
...
-    "index.js": [
-      "prettier --write",
-      "git add"
-    ],
-    "**/*.jsx": [
-      "prettier --write",
-      "git add"
-    ]
+    "index.js": ["prettier --write", "git add"],
+    "**/*.jsx": ["prettier --write", "git add"]
```
filoozom: Was this changed by `npm run prettier:write`?

prateekbh (approved):

TextField/TextField.jsx
```diff
@@ -75,6 +76,19 @@ class TextFieldInput extends MaterialComponent {
   materialDom(allprops) {
     let { className, ...props } = allprops;
```
prateekbh: @mellogarrett can we just do `className = className || ''`.
This will clean up nested conditions

mellogarrett: I don't know how I would write that in this case... `className = className + "..." || "..."` doesn't work

prateekbh: just `className = className || ''`, below line 76 doesn't work?
className should be a simple string isnt it?

mellogarrett: It is a string, but if it exists we want to add to it. If it's null or undefined we want to set it. Right?

prateekbh: so the code above will just initialize an empty string if its null or will leave it as is if its there

mellogarrett: Oh, duh... sorry, I get it now.

mellogarrett: I was thinking do it all on one line

Issue #445, opened 2017-12-06
Select aria-selected not updating correctly


mellogarrett:
If you pass `selectedIndex=null` to a Select component, it does not set all of its items to `aria-selected: false` like you would expect.

prateekbh:
ohh nice catch

mellogarrett:
This has been fixed in #447 

`material-components-web` did not like having `selectedIndex` set to null. It updates the aria-selected attr correctly, however, when you set it to -1 instead.

PR #447, opened 2017-12-06
Select - fix "unselected" state

mellogarrett:
This addresses an issue where passing -1 as `selectedIndex` would cause `hintText` to be hidden. The preferred method to _clear_ a selection is now to pass `selectedIndex={1}` instead of `selectedIndex={null}`. 

`selectedIndex={null}` leaves `aria-selected=true`

### DIFF
--- a/Select/Select.jsx
+++ b/Select/Select.jsx
@@ -43,7 +43,10 @@ class Select extends MaterialComponent {
     this.updateSelection();
     if (this.MDComponent && this.MDComponent.foundation_) {
       this.MDComponent.foundation_.resize();
-      if (this.props.selectedIndex == null || this.props.selectedIndex == -1) {
+      if (
+        this.props.selectedIndex === null ||
+        this.props.selectedIndex === -1
+      ) {
         this.MDComponent.foundation_.adapter_.setSelectedTextContent(
           this.props.hintText
         );
--- a/Select/Select.jsx
+++ b/Select/Select.jsx
@@ -43,6 +43,11 @@ class Select extends MaterialComponent {
     this.updateSelection();
     if (this.MDComponent && this.MDComponent.foundation_) {
       this.MDComponent.foundation_.resize();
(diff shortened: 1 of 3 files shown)


Select/Select.jsx
```diff
@@ -43,6 +43,11 @@ class Select extends MaterialComponent {
     this.updateSelection();
     if (this.MDComponent && this.MDComponent.foundation_) {
       this.MDComponent.foundation_.resize();
+      if (this.props.selectedIndex == null || this.props.selectedIndex == -1) {
```
filoozom: Can we use `===` here instead of `==`?

filoozom:
Damn, that pesky `prettier` bug again... Really need to find a solution for this!

mellogarrett:
Okay, `prettier --write` is what changed my docs page to make it fail. And I can't `git commit` without running `prettier --write` now???

mellogarrett:
I had to `git commit -n` to get the correct format to commit :/

filoozom:
`git commit` without `prettier --write` should never be possible, but after `prettier --write`, `prettier -l` should not return an error. So that's strange.

mellogarrett:
Not sure why, but what `prettier --write` created for me locally did not pass the Travis test.

mellogarrett:
Fixes #445

filoozom:
Do you have `prettier` installed globally or something? Maybe a version mismatch?

mellogarrett:
Not global. And I did a `rm -r node_modules && npm i` to make sure I had the _correct_ version

mellogarrett:
@prateekbh any reason this hasn't been merged?

prateekbh:
nups just me moving to a new country and fighting jetlag.
also please feel free to merge if approved.

mellogarrett:
haha sorry, wasn't trying to be pushy. 

Congrats again on the move!!

PR #474, opened 2017-12-19
Typescript declaration improvements

cvuorinen:
Some typing fixes / improvements & also some fixes to docs and samples (I'm going through the docs and testing the typings with the samples so found some issues there).

Prop `disabled?: boolean` is redundant, since it's inherited from HTMLAttributes.

There are more coming later, but started this PR already so you can see some progress.

ping @YePpHa can you check this out also?

### DIFF
--- a/Icon/index.d.ts
+++ b/Icon/index.d.ts
@@ -1,5 +1,4 @@
 import MaterialComponent from '../MaterialComponent';
 import { VNode } from 'preact';
 
-declare interface IIconProps extends JSX.HTMLAttributes {}
-export default class Icon extends MaterialComponent<IIconProps, {}> {}
+export default class Icon extends MaterialComponent<JSX.HTMLAttributes, {}> {}
--- a/Icon/index.d.ts
+++ b/Icon/index.d.ts
@@ -1,7 +1,5 @@
 import MaterialComponent from '../MaterialComponent';
 import { VNode } from 'preact';
 
-declare interface IIconProps extends JSX.HTMLAttributes {
-  disabled?: boolean;
-}
+declare interface IIconProps extends JSX.HTMLAttributes {}
 export default class Icon extends MaterialComponent<IIconProps, {}> {}
--- a/IconToggle/index.d.ts
+++ b/IconToggle/index.d.ts
@@ -2,10 +2,15 @@ import MaterialComponent from '../MaterialComponent';
 import { VNode } from 'preact';
 import { MDCFoundation, MDCComponent, MDCRipple } from '../MaterialComponentsWeb';
 
+declare interface IconToggleState {
+  label?: string;
+  content?: string;
+  cssClass?: string;
+}
+
 declare interface IIconToggleProps extends JSX.HTMLAttributes {
-  disabled?: boolean;
-  'data-toggle-on'?: any;
-  'data-toggle-off'?: any;
+  'data-toggle-on'?: IconToggleState;
+  'data-toggle-off'?: IconToggleState;
 }
 export default class IconToggle extends MaterialComponent<IIconToggleProps, {}> {
   MDComponent: MDCIconToggle;
--- a/IconToggle/index.d.ts
+++ b/IconToggle/index.d.ts
@@ -2,15 +2,15 @@ import MaterialComponent from '../MaterialComponent';
 import { VNode } from 'preact';
 import { MDCFoundation, MDCComponent, MDCRipple } from '../MaterialComponentsWeb';
 
-declare interface IconToggleState {
+type IconToggleData = {
   label?: string;
   content?: string;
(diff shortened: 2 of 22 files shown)


IconToggle/index.d.ts
```diff
@@ -2,10 +2,15 @@ import MaterialComponent from '../MaterialComponent';
...
+}
+
 declare interface IIconToggleProps extends JSX.HTMLAttributes {
-  disabled?: boolean;
-  'data-toggle-on'?: any;
-  'data-toggle-off'?: any;
+  'data-toggle-on'?: IconToggleState;
+  'data-toggle-off'?: IconToggleState;
 }
 export default class IconToggle extends MaterialComponent<IIconToggleProps, {}> {
```
YePpHa: `IconToggleState` is not added to the MaterialComponent.

```TypeScript
export default class IconToggle extends MaterialComponent<IIconToggleProps, IconToggleState> {
```

IconToggle/index.d.ts
```diff
@@ -2,10 +2,15 @@ import MaterialComponent from '../MaterialComponent';
 import { VNode } from 'preact';
 import { MDCFoundation, MDCComponent, MDCRipple } from '../MaterialComponentsWeb';
 
+declare interface IconToggleState {
```
YePpHa: I'm not sure if we want to keep it consistent to prepend an `I` for interface like I've done for everything else.

IconToggle/index.d.ts
```diff
@@ -2,10 +2,15 @@ import MaterialComponent from '../MaterialComponent';
...
+  label?: string;
+  content?: string;
+  cssClass?: string;
+}
+
 declare interface IIconToggleProps extends JSX.HTMLAttributes {
```
YePpHa: Missing the prop `onChange`.

IconToggle/index.d.ts
cvuorinen: Ah, yes. Confusing name for the type. It's not the State for the component. Just named it after something discussed in the docs, but it's obviously misleading in this context. I'll rename it to something like IconToggleData etc.

IconToggle/index.d.ts
cvuorinen: That is also inherited from `HTMLAttributes` already

porfirioribeiro:
Cool @cvuorinen ! Nice work, thank's!

cvuorinen:
I added rest of the fixes.

If you can check the typings @YePpHa & @porfirioribeiro that would be great.

Also, @prateekbh could you check the docs & samples I updated if all those changes look valid? It's a bit confusing trying to figure out what MDC version is using what props etc...

Icon/index.d.ts
```diff
@@ -1,7 +1,5 @@
 import MaterialComponent from '../MaterialComponent';
 import { VNode } from 'preact';
 
-declare interface IIconProps extends JSX.HTMLAttributes {
-  disabled?: boolean;
-}
-export default class Icon extends MaterialComponent<IIconProps, {}> {}
\ No newline at end of file
+declare interface IIconProps extends JSX.HTMLAttributes {}
```
YePpHa: If we just have an empty interface just use `JSX.HTMLAttributes` directly.

List/index.d.ts
```diff
@@ -29,7 +29,11 @@ declare class ItemIcon extends MaterialComponent<IItemIconProps, {}> {
   getProxyClassName(props: IItemIconProps): string;
 }
 
-declare class ItemAvatar extends MaterialComponent<JSX.HTMLAttributes, {}> {}
+declare interface IItemAvatarProps extends JSX.HTMLAttributes {
+  'start-detail'?: boolean;
+  'end-detail'?: boolean;
+}
+declare class ItemAvatar extends MaterialComponent<IItemAvatarProps, {}> {}
```
YePpHa: `ItemAvatar` extends from `ItemIcon`. Let's just extend it directly. There are also no props from `ItemAvatar` other than the inherited props.

If in the future we want more props on `ItemAvatar` we can do as `LinkItem`.

cvuorinen:
Thanks for the review @YePpHa, I made the changes.

prateekbh:
Honestly I am kinda noob in TS so please feel free to LG and merge when you guys see fit... and open an issue for release. 
We have on release waiting for `mdc-web 0.0.27`

cvuorinen:
@prateekbh I asked if you could check the docs and samples as I also updated some of those ;)
E.g. https://github.com/prateekbh/preact-material-components/pull/474/files#diff-3012bf945b75715c05fc7a582f5e1e6cL27

prateekbh:
@cvuorinen woops missed that.
Feel free to merge once @YePpHa LGs

PR #475, opened 2017-12-19
changes from 0.0.27

prateekbh:
- fixes new `select` dom
- fixes dom of list


mellogarrett:
Did `material-components-web` change the `Select` syntax??

Issue #476, opened 2017-12-20
Keep consistency in listeners to always pass an event object as an argument


YePpHa:
I noticed in [PR/474](https://github.com/prateekbh/preact-material-components/pull/474) that some of the listeners don't get the event object passed as an argument (see [/Menu/Menu.jsx#L43](https://github.com/prateekbh/preact-material-components/blob/master/Menu/Menu.jsx#L43) and [/Dialog/Dialog.jsx#L26](https://github.com/prateekbh/preact-material-components/blob/master/Dialog/Dialog.jsx#L26)).

We should keep this consistent to always pass the event object.

PR #482, opened 2018-01-02
Made all event listeners consistent to always return Event

YePpHa:
Resolves: #476.

All the event listeners will now pass the Event object as an argument to make it consistent for all the event listeners.

__Breaking change__
The events `onChange` and `onInput` will pass the event object instead of the value from `MDCComponent`. To still allow easy access to the value of the slider I've added the method `getValue`.

### DIFF
--- a/Slider/Slider.jsx
+++ b/Slider/Slider.jsx
@@ -13,27 +13,36 @@ export default class Slider extends MaterialComponent {
     this._onChange = this._onChange.bind(this);
     this._onInput = this._onInput.bind(this);
   }
-  _onChange() {
+
+  _onChange(e) {
     if (this.props.onChange) {
-      this.props.onChange(this.MDComponent.value);
+      this.props.onChange(e);
     }
   }
-  _onInput() {
+
+  _onInput(e) {
     if (this.props.onInput) {
-      this.props.onInput(this.MDComponent.value);
+      this.props.onInput(e);
     }
   }
+
   componentDidMount() {
     this.MDComponent = new MDCSlider(this.base);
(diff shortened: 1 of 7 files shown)


Slider/Slider.jsx
```diff
@@ -13,27 +13,36 @@ export default class Slider extends MaterialComponent {
     this._onChange = this._onChange.bind(this);
     this._onInput = this._onInput.bind(this);
   }
-  _onChange() {
+
+  _onChange(e) {
     if (this.props.onChange) {
-      this.props.onChange(this.MDComponent.value);
+      this.props.onChange(e);
```
prateekbh: can you please confirm that some value is passed here?
last time i check nothing was passed hence, this was passed

Slider/Slider.jsx
```diff
@@ -13,27 +13,36 @@ export default class Slider extends MaterialComponent {
...
     }
   }
-  _onInput() {
+
+  _onInput(e) {
     if (this.props.onInput) {
-      this.props.onInput(this.MDComponent.value);
+      this.props.onInput(e);
```
prateekbh: same as above

Slider/Slider.jsx
YePpHa: The MDCComponent `listen` method attaches an event listener on the DOM element (see [packages/mdc-base/component.js#L89](https://github.com/material-components/material-components-web/blob/master/packages/mdc-base/component.js#L89)).

In the image below is an example of the output of both `onInput` and `onChange`. This is the code that outputs to the console: `console.log(evt, slider.getValue());`.
image

Issue #515, opened 2018-01-14
primary/accent as boolean props?


developit:
Just a thought - rather than having to do `<Button className="mdc-theme--primary-bg">` could we set up `<Button primary>`?

mellogarrett:
Sounds like a great idea! Maybe accept `<Button secondary>` _and_ `<Button accent>`

prateekbh:
IIUC `primary` and `accent` are reserved terminology in case of material design. let use that and put the apt class name in if we have an agreement.

PR #517, opened 2018-01-18
Added "primary", "secondary", and "accent" props to <Button>

mellogarrett:
These props set the background color of the `Button` by adding the appropriate `mdc-theme--[prop]-bg` class

### DIFF
--- a/Button/Button.jsx
+++ b/Button/Button.jsx
@@ -15,15 +15,21 @@ class Button extends MaterialComponent {
     super();
     this.componentName = "button";
     this._mdcProps = ["dense", "raised", "compact", "unelevated", "stroked"];
+    this.themeProps = ["primary", "secondary", "accent"];
   }
   componentDidMount() {
     super.attachRipple();
   }
   materialDom(props) {
     const ButtonElement = props.href ? "a" : "button";
+    let className = "";
+    this.themeProps.forEach(themeProp => {
+      if (themeProp in props) className += _generateThemeClass(themeProp);
+    });
 
     return (
       <ButtonElement
+        className={className}
         ref={control => {
           this.control = control;
         }}
@@ -41,6 +47,15 @@ class ButtonIcon extends Icon {
     this.componentName = "button__icon";
   }
 }
+
+const _generateThemeClass = prop => {
+  const mappings = {
+    primary: "primary",
+    secondary: "secondary",
+    accent: "secondary"
+  };
+  return `mdc-theme--${mappings[prop]}-bg `;
+};
 
 Button.Icon = ButtonIcon;
 export default Button;
--- a/sample/home.jsx
+++ b/sample/home.jsx
@@ -86,9 +86,9 @@ export default class Home extends Component {
         </Drawer.TemporaryDrawer>
 
         <Button
-          raised={true}
-          accent={true}
-          ripple={true}
+          primary
+          raised
+          ripple
           onClick={e => {
             this.drawer.MDComponent.open = true;
           }}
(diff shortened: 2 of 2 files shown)


mellogarrett:
Fixes #515

Button/Button.jsx
```diff
@@ -42,5 +48,14 @@ class ButtonIcon extends Icon {
   }
 }
 
+const _generateThemeClass = prop => {
```
prateekbh: i guess there are more components that can use this e.g. fab.

shall we move this into a util instead?

Button/Button.jsx
```diff
@@ -15,15 +15,21 @@ class Button extends MaterialComponent {
     super();
     this.componentName = "button";
     this._mdcProps = ["dense", "raised", "compact", "unelevated", "stroked"];
+    this.themeProps = ["primary", "secondary", "accent"];
```
prateekbh: accent is no more there only primary/secondary

Button/Button.jsx
mellogarrett: Just create a `utils` folder for it?

prateekbh: `utils.js/ThemeUtils.js` may be.

utils/generateThemeClass.js
```diff
@@ -0,0 +1 @@
+export default prop => `mdc-theme--${prop}-bg `;
```
prateekbh: can we call it ThemeUtils? so that we can put more stuff in here in future?

mellogarrett: Capital `T` or lowercase `t`? I thought the precedent was for capitals to be reserved for components.

prateekbh: ohyeah u r right. themeUtils.js

Button/Button.jsx
```diff
@@ -15,15 +16,21 @@ class Button extends MaterialComponent {
...
+    this.themeProps = ["primary", "secondary"];
   }
   componentDidMount() {
     super.attachRipple();
   }
   materialDom(props) {
     const ButtonElement = props.href ? "a" : "button";
+    let className = "";
+    this.themeProps.forEach(themeProp => {
+      if (themeProp in props) className += generateThemeClass(themeProp);
```
YePpHa: If you set an attribute on an element in preact (using Typescript) without defining its value it would default to `true` if the type is a boolean. I would suggest making a check to verify its value is not `false`. Someone might have some code that only sets a button to primary/secondary on certain conditions (see example below). Adding the check should be fairly simple: `if (themeProp in props && props[themeProp] !== false) ...`

__Example:__
```Typescript
const isPrimary = someRandomFunc();

<Button primary={isPrimary}>My button</Button>
```

Button/Button.jsx
```diff
@@ -15,15 +16,22 @@ class Button extends MaterialComponent {
...
+    this.themeProps = ["primary", "secondary"];
   }
   componentDidMount() {
     super.attachRipple();
   }
   materialDom(props) {
     const ButtonElement = props.href ? "a" : "button";
+    let className = "";
+    this.themeProps.forEach(themeProp => {
+      if (themeProp in props && themeProp !== false)
```
YePpHa: You forgot that `themeProp` is a possible key of a prop in `props`. It's also a string and it wouldn't make sense to compare a string to a boolean. Change `themeProp !== false` to `props[themeProp] !== false`.

mellogarrett: Whoops!

themeUtils/generateThemeClass.js
```diff
@@ -0,0 +1 @@
+export default prop => `mdc-theme--${prop}-bg `;
```
YePpHa: @prateekbh should we keep a space here? Shouldn't it be handled where it's used?

prateekbh: makes sense

YePpHa: Remove the space and move it to where it's called.

PR #518, opened 2018-01-18
fixing list

prateekbh:
fixes class names in list


prateekbh:
adding specially @YePpHa for the TS review

Issue #519, opened 2018-01-22
Added "primary", "secondary" props to <Fab>


prateekbh:
Check if fab still supports this and add it back

PR #554, opened 2018-02-05
Added "primary" & "secondary" props to Fab components

mellogarrett:


mellogarrett:
Addresses #519

Fab/Fab.jsx
```diff
@@ -11,13 +12,23 @@ class Fab extends MaterialComponent {
     super();
     this.componentName = "fab";
     this._mdcProps = ["mini", "exited"];
+    this.themeProps = ["primary", "secondary"];
   }
   componentDidMount() {
     super.attachRipple();
   }
   materialDom(props) {
+    let className = "";
```
prateekbh: `= props.class || props.className`

prateekbh: also to my horror we missed it in button as well, can you add it there as well?

YePpHa: It should actually be `= props.class || props.className || ""` because it would otherwise be undefined if `class` or `className` are not defined

mellogarrett: And then I would need to move `{...props}` before `class={className}`, correct?

YePpHa: We should actually omit `class` and `className` from {...props} but the order doesn't really matter. See https://github.com/prateekbh/preact-material-components/blob/master/Select/Select.jsx#L112 for how to omit the properties (please note that I clone the props as to be certain that I don't modify the actual props).

Fab/Fab.jsx
```diff
@@ -11,13 +12,23 @@ class Fab extends MaterialComponent {
     super();
     this.componentName = "fab";
     this._mdcProps = ["mini", "exited"];
+    this.themeProps = ["primary", "secondary"];
   }
   componentDidMount() {
     super.attachRipple();
   }
   materialDom(props) {
+    let className = props.class || props.className || "";
```
YePpHa: We should omit `class` and `className` from `props`. See https://github.com/prateekbh/preact-material-components/blob/master/Select/Select.jsx#L112 for how to omit the properties (please note that I clone the props as to be certain that I don't modify the actual props). Otherwise `class` is not being set at all (I tested this).

mellogarrett: This works if you pass `className` to Fab instead of `class`  🤔

prateekbh: not sure why, it should treat `class` as a normal prop however

mellogarrett: Yea its being stripped out of `props` for some reason

YePpHa: I found the culprit https://github.com/prateekbh/preact-material-components/blob/master/MaterialComponent.js#L64

prateekbh: oh heck! yeah i did that to make sure we do not override class accidentally just like this.

mellogarrett: So this is desired behavior then? Intended use is to just pass `className` prop?

prateekbh: yeah just use `className`. I'll make sure that this gets merged properly

YePpHa: It still doesn't use the `class` prop as it gets deleted by the above mentioned code. If we remove `delete componentProps.class;` it will add the user-specified classes correctly.

YePpHa: However, I am not certain what side effects that might have on the other components.

YePpHa:
@prateekbh is it supposed to be that the user needs to use `className` instead of `class`?

prateekbh:
@YePpHa As i see it for now, yes thats a presumption

YePpHa:
Okay, I will approve this then.

Issue #585, opened 2018-02-08
Add chips component


prateekbh:
https://github.com/material-components/material-components-web/tree/master/packages/mdc-chips

YePpHa:
I can implement this.

YePpHa:
@prateekbh I got a question regarding on how you want me to export the two classes Chip and ChipSet. The problem is that the package is named Chips and therefore it doesn't make sense to export Chip as default.
Should I export it the same as with the Drawers or should I export them normally: `import { Chip, ChipSet } from '/Chips'`?

prateekbh:
I guess lets just name is `Chips`

and let it hide class `chip-set` from user. User can use `Chips.Chip` for implementation

```
import Chips from 'pmc/Chips'
.
.
.
<Chips>
<Chips.Chip>...</>
<Chips.Chip>...</>
<Chips.Chip>...</>
</Chips>
```

YePpHa:
Okay

PR #589, opened 2018-02-08
feat(Chips): implement chips

YePpHa:
Issue: #585

I've implemented the `Chip` and `ChipSet` component. Currently there's not a lot to them, but I imagine they will be extended upon in the future.

### DIFF
--- a/Chips/Chips.jsx
+++ b/Chips/Chips.jsx
@@ -36,16 +36,24 @@ class Chip extends MaterialComponent {
   }
 
   materialDom(props) {
-    return (
-      <div>
-        <div class="mdc-chip__text">{props.children}</div>
-      </div>
-    );
+    return <div>{props.children}</div>;
+  }
+}
+
+class ChipText extends MaterialComponent {
+  constructor() {
+    super();
+    this.componentName = "chip__text";
+  }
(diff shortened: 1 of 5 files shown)


YePpHa:
@mellogarrett problem is that https://github.com/material-components/material-components-web haven't implemented that part yet. However, I can see that they have two PRs that extend upon the current Chips: [feat(chips): Handle leading/trailing icon styles](https://github.com/material-components/material-components-web/pull/2191) and [feat(chips): Add `mdc-chip-set--choice` variant](https://github.com/material-components/material-components-web/pull/2215)

Chips/Chips.jsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  componentWillUnmount() {
+    this.MDComponent.destroy && this.MDComponent.destroy();
+  }
+
+  materialDom(props) {
+    return (
+      <div>
+        <div class="mdc-chip__text">{props.children}</div>
```
prateekbh: Actually lets make this a different component ChipText as we know soon ChipIcon will come in too. and if it extends any further this will be an unnecessary breaking change.
WDYT?

PR #592, opened 2018-02-15
Tabs active index

### LINKED ISSUE #550: Add activeIndex as a prop to Tabs
Currently `active` is a prop to every `tab` in `tabs`. 

a.) This can lead to issues if a dev give multiple `active` classes.

b.) changing just this prop, does not update the indicator.

Hence would be best to use this in `tabs` and call underlying function on `MDComponent`

prateekbh:
Fixes #550


Tabs/Tabs.jsx
```diff
@@ -21,10 +29,14 @@ class Tabs extends MaterialComponent {
   }
   componentDidMount() {
     this.MDComponent = new MDCTabBar(this.control);
+    setActiveTabIndex(defaultProps, this.props, this.MDComponent);
```
mellogarrett: This ignores the `activeTabIndex` prop on initial mount and only uses the prop on update

mellogarrett: This ignores the `activeTabIndex` prop on initial mount and only uses the prop on update

prateekbh: fixed

prateekbh:
@mellogarrett please review

PR #614, opened 2018-02-21
adding line-ripple

prateekbh:
Adding LineRipple as a component


package.json
```diff
@@ -71,6 +71,7 @@
     "preact": "*"
   },
   "devDependencies": {
+    "@material/chips": "^0.31.0",
```
YePpHa: Shouldn't this be in `dependencies` and not `devDependencies`?

prateekbh: oh yes

prateekbh:
@YePpHa  done!

LineRipple/index.d.ts
```diff
@@ -0,0 +1,4 @@
+import MaterialComponent from '../MaterialComponent';
+import { VNode } from 'preact';
+
+export default class LineRipple extends MaterialComponent<JSX.HTMLAttributes, {}> {}
```
YePpHa: Perhaps it would be an idea to define the MDCLineRipple (property `MDComponent`) too (see https://github.com/prateekbh/preact-material-components/blob/master/Select/index.d.ts#L23 for how to define it and https://github.com/material-components/material-components-web/tree/master/packages/mdc-line-ripple#mdclineripple for the definition of the class).

prateekbh: oohhhh I didnt know about this

prateekbh:
@YePpHa Done V2

LineRipple/index.d.ts
```diff
@@ -1,4 +1,20 @@
...
+  activate(): void;
+  deactivate(): void;
+  setRippleCenter(xCoordinate: number): void;
+}
+
+declare class MDCLineRipple extends MDCComponent<MDCLineRippleFoundation> {
+  activate(): void;
+  deactivate(): void;
+  setRippleCenter(xCoordinate: number): void;
+  handleTransitionEnd(e: Event): void;
```
YePpHa: This is supposed to be on the foundation class.

prateekbh:
Corrected

Issue #612, opened 2018-02-20
Implement Chips Icon


YePpHa:
I've started work [here](https://github.com/YePpHa/preact-material-components/tree/feat/chips/leading-icon) but it still needs the new version of mdc.

prateekbh:
@YePpHa can we add this in 1.3.6?

YePpHa:
@prateekbh sure, I will make a PR when I get home

prateekbh:
@YePpHa can you please add the docs page as well with this! so we'll be almost good to publish

YePpHa:
@prateekbh sure I will start work on it

PR #619, opened 2018-02-24
feat(chips): implement leading and trailing icon

YePpHa:
3 days late PR (forgot to submit it) - #612


Chips/Chips.jsx
```diff
@@ -40,6 +40,18 @@ class Chip extends MaterialComponent {
   }
 }
 
+class ChipIcon extends MaterialComponent {
```
prateekbh: this should extend `Icon`.
It brings `material-icons` classes with it

https://github.com/material-components/material-components-web/tree/master/packages/mdc-chips#trailing-icon

PR #624, opened 2018-03-01
Class name fix

prateekbh:
- Fixes className bug.
- No component will have to handle user given class names.
- Adds visual diff testing via pupeteer.
- Docs will now be deployed from release branch, which means code change and doc change can be PRed simultaneously.


MaterialComponent.js
```diff
@@ -25,9 +25,12 @@ export default class MaterialComponent extends Component {
       MDCRipple.attachTo(this.control);
     }
   }
-  // Build the className
+  // Build the className based on component names and mdc props
   buildClassName(props) {
```
YePpHa: The `props` parameter is not used at all even when calling `buildClassName()`.

MaterialComponent.js
```diff
@@ -53,23 +57,30 @@ export default class MaterialComponent extends Component {
...
   materialDom(props) {
     return h("div", Object.assign({}, props), props.children);
   }
+
   render() {
     this.buildClassName();
     // Fetch a VNode
     const componentProps = this.props;
+    const userDefinedClasses =
+      componentProps.className || componentProps.class || "";
```
YePpHa: Might be an idea to actually merge them instead.

```JavaScript
const userDefinedClasses =
  ((componentProps.className || "") + " " + (componentProps.class || ""))
  .split(' ') // split the class names into an array
  .filter((value, index, self) => self.indexOf(value) === index && value !== "") // Unique + exclude empty class names
  .join(' '); // Merge them into a string again
```

MaterialComponent.js
```diff
@@ -53,23 +57,30 @@ export default class MaterialComponent extends Component {
...
     if (componentProps.class) {
```
YePpHa: If we check for whether `class` exists why do we also remove `className`? Shouldn't it be put into its own if statement?

```JavaScript
// Short example
if (componentProps.class) delete componentProps.class;
if (componentProps.className) delete componentProps.className;
```

MaterialComponent.js
```diff
@@ -53,23 +57,30 @@ export default class MaterialComponent extends Component {
...
+      delete componentProps.className;
       delete componentProps.class;
     }
     const element = this.materialDom(componentProps);
     element.attributes = element.attributes || {};
-    // Fix for className
-    element.attributes.class = this.getClassName(element);
-    // element.attributes.className = this.getClassName(element);
+
+    element.attributes.className = `${userDefinedClasses} ${this.getClassName(
```
YePpHa: The `this.getClassName(element)` should be included in the logic I included in my review of the `userDefinedClasses` part. This will make sure that we don't have duplicate class names.

```JavaScript
const userDefinedClasses =
  ((componentProps.className || "") + " " + (componentProps.class || ""));
...
element.attributes.className =
  (userDefinedClasses + " " + this.getClassName(element))
  .split(' ') // split the class names into an array
  .filter((value, index, self) => self.indexOf(value) === index && value !== "") // Unique + exclude empty class names
  .join(' '); // Merge them into a string again;
```

YePpHa:
I just have a question related to the screenshot test. Wouldn't the animated components be inconsistent (ex. docs/tests/golden/wide/component/linear-progress.png)?

prateekbh:
@YePpHa yep, that why I am skipping `linear-progress` for now.
Will look for something. But till then, we have some testing in place

MaterialComponent.js
prateekbh: wondering if looking for unique classNames become too much for every render? WDYT?

YePpHa: I don't think it would be a problem except if you have a ludicrous amount of class names and you call `render()` every second, but then I would think there are other problems that would occur before this becomes an issue (e.g. the DOM rendering has problems keeping up).

Breakdown of the methods used:
- `join` should be fairly inexpensive.
- `split` should also be inexpensive.
- `filter` is just a for loop that creates a new array with a condition. Highly doubt this would be an issue at all.
- `indexOf` should be fine too (https://jsperf.com/array-indexof-vs-set-has) but is probably the one that is the most expensive of them all.

prateekbh: Done

MaterialComponent.js
```diff
@@ -53,23 +57,34 @@ export default class MaterialComponent extends Component {
...
   render() {
     this.buildClassName();
     // Fetch a VNode
     const componentProps = this.props;
-    if (componentProps.class) {
-      // We delete class prop here so that any sub node's class doesn't get over-ridden from this
-      delete componentProps.class;
-    }
+    const userDefinedClasses =
+      componentProps.className || componentProps.class || "";
```
YePpHa: Change it to `((componentProps.className || "") + " " + (componentProps.class || ""))` to allow for having both `className` and `class`.

MaterialComponent.js
```diff
@@ -53,23 +57,34 @@ export default class MaterialComponent extends Component {
...
+    // We delete class props and add them later in the final
+    // step so every component does not need to handle user specified classes.
+    if (componentProps.class) delete componentProps.class;
```
YePpHa: Might be better to not reference `class` as a dot notation as it's a [reserved word](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords). I highly doubt it would cause issues, but we sometimes have something called IE and it has its quirks.

```JavaScript
if (componentProps['class']) delete componentProps['class'];
```

docs/package.json
```diff
@@ -4,12 +4,12 @@
...
     "start": "if-env NODE_ENV=production && npm run -s serve || npm run -s dev",
-    "build": "preact build --no-prerender",
-    "build:prod": "preact build --template template.html --no-prerender && npm run copyAssets",
+    "build": "preact build",
+    "build:prod": "preact build --template template.html && npm run copyAssets",
     "copyAssets": "cp -R images *.png *.json build/",
     "serve": "preact build && preact serve",
     "dev": "preact watch",
-    "test": "npm run prettier",
+    "test": "npm run build:prod && ../node_modules/.bin/mocha ./tests/visdiff.js --timeout 5000 --exit",
```
filoozom: Would be nice to keep `prettier` in the `test` script.

scripts/travis.sh
```diff
@@ -5,8 +5,11 @@ SUBFOLDER=$1
 
 function test {
     set -ev
+    npm install
+    npm run build
```
filoozom: Shouldn't this be included in `docs/package.json`? Guess in the build step. Because either way we can't build docs without running `npm install` on the parent directory. Also, this forces two builds when we're just testing the library and not the docs.

scripts/travis.sh
```diff
@@ -5,8 +5,11 @@ SUBFOLDER=$1
...
     cd $TRAVIS_BUILD_DIR/$SUBFOLDER
     npm install
+    pwd
```
filoozom: Guess that was for testing, but isn't necessary anymore? 😉

docs/tests/visdiff.js
```diff
@@ -0,0 +1,188 @@
...
+    if (!fs.existsSync(`${testDir}/wide`)) fs.mkdirSync(`${testDir}/wide`);
+    if (!fs.existsSync(`${testDir}/wide/component`)) {
+      fs.mkdirSync(`${testDir}/wide/component`);
+    }
+    browser = await puppeteer.launch();
+    page = await browser.newPage();
+  });
+
+  after(() => {
+    server.stop();
```
filoozom: Add `browser.close()` here, then `--exit` on `mocha` in `package.json` shouldn't be required (but still have to test that one out).

docs/tests/visdiff.js
```diff
@@ -0,0 +1,188 @@
...
+}
+
+function compareScreenshots(fileName) {
+  return new Promise((resolve, reject) => {
+    const img1 = fs.createReadStream(`${testDir}/${fileName}.png`).pipe(new PNG()).on('parsed', doneReading);
+    const img2 = fs.createReadStream(`${goldenDir}/${fileName}.png`).pipe(new PNG()).on('parsed', doneReading);
+
+    console.log(base64Img.base64Sync(`${testDir}/${fileName}.png`));
+    console.log("______________________");
+    console.log(base64Img.base64Sync(`${goldenDir}/${fileName}.png`));
```
filoozom: Printing `base64` to console kills Travis and makes it quite unreadable.

docs/package.json
```diff
@@ -4,12 +4,12 @@
   "description": "preact material component's site",
   "scripts": {
     "start": "if-env NODE_ENV=production && npm run -s serve || npm run -s dev",
-    "build": "preact build --no-prerender",
-    "build:prod": "preact build --template template.html --no-prerender && npm run copyAssets",
+    "build": "preact build",
+    "build:prod": "preact build --template template.html && npm run copyAssets",
     "copyAssets": "cp -R images *.png *.json build/",
```
filoozom: Would be nice to replace `cp` with something that would work on Windows too. Maybe `cp-cli`? Seems pretty old and not that popular though.

MaterialComponent.js
prateekbh: both at the same time! ppl do that?

docs/package.json
prateekbh: this one's a big change, plz open a different issue. will take this up later

docs/tests/visdiff.js
prateekbh: oh no no this will be reverted, its just a temporary thing

docs/tests/visdiff.js
prateekbh: oh TIL

scripts/travis.sh
prateekbh: yeah stuff is lil bit messed up right now

docs/package.json
filoozom: Sure. I actually changed it in my tests (#626) because I only have a Windows computer on hand right now.

.travis.yml
```diff
@@ -24,11 +25,15 @@ deploy:
...
 # blocklist
 branches:
   except:
-    - gh-pages
\ No newline at end of file
+    - gh-pages
+
+sudo: required
+addons:
+ chrome: stable
```
filoozom: Is this necessary? `puppeteer` should install chrome by itself.

prateekbh: nups without this, it doesnt work

MaterialComponent.js
prateekbh: done

docs/package.json
prateekbh: done

scripts/travis.sh
prateekbh: its apparently not doing for parent folder from the child folder

filoozom:
Does it really make sense to screenshot whole documentation pages instead of single components? This way we need to update screenshots for every single tiny change, like typos etc, which will make it a lot more complicated to maintain.

prateekbh:
@filoozom we can specify some level of threshold which will be fine with typos and all

prateekbh:
@YePpHa @filoozom if this looks fine, i'll probably force merge this and apply @filoozom 's fix on top.

Please review

scripts/travis.sh
```diff
@@ -5,8 +5,9 @@ SUBFOLDER=$1
 
 function test {
     set -ev
-    cd $TRAVIS_BUILD_DIR/$SUBFOLDER
     npm install
+    npm run build
+    cd $TRAVIS_BUILD_DIR/$SUBFOLDER
     npm test
```
filoozom: I think there's a `npm install` missing here after the `cd`. But that will change in #626 anyway.

MaterialComponent.js
YePpHa: Not sure why they would do that. I only use `class` but people are stupid.

developit:
Just hit this and updating fixed it! <3

PR #626, opened 2018-03-02
Upload snapshots to GH after test

filoozom:
Don't mind me, just trying out a few Travis builds.

PR #628, opened 2018-03-03
[WIP] Post comment containing failed pictures to PR

filoozom:
The goal of this bot is to make it easier to update visual diff pictures created in #624 and #626.

- Creates a Gist
- Pushes a zip file to it (can't do it in one step sadly)
- Posts a comment on the failed PR with a link to the zip

The reason why we use Gist is that I don't think that it's currently possible to directly create a comment with files with the GitHub API.

**WIP**: Figure out Travis secrets

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -166,16 +166,6 @@
       "integrity": "sha512-z55ocwKBRLryBs394Sm3ushTtBeg6VAeuku7utSoSnsJKvKcnXFIyC6vh27n3rXyxSgkJBBCAvyOn7gSUcTYjg==",
       "dev": true
     },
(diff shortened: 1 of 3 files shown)


filoozom:
Ok, so this isn't actually doable without an external server I think, so I'll just put this on hold for a bit until I find a solution.

developit:
This would be an awesome dedicated service

filoozom:
Was thinking about that too actually. Do you have any suggestions / whishes? I'm mostly stuck on figuring out the security aspect of it right now.

prateekbh:
so something exactly like percy!

filoozom:
Closing in favour of #668

PR #630, opened 2018-03-05
Consolidate ref setting

developit:
Consolidate ref setting for this.control into a shared method to avoid closures in render.


YePpHa:
Just use `this.base` as it always refers the the base element.

developit:
@YePpHa That's an option too (though technically not part of Preact's documented API and refs are preferred).

PR #745, opened 2018-04-11
Add disabled class to TextField

GusRuss89:
Disabled text fields are actually supposed to have an extra class on the field wrapper. This PR just adds `mdc-text-field--disabled` class if `disabled` prop is truthy.

### DIFF
--- a/TextField/TextField.jsx
+++ b/TextField/TextField.jsx
@@ -85,6 +85,10 @@ class TextFieldInput extends MaterialComponent {
     if ('trailingIcon' in props) {
       className += ' mdc-text-field--box mdc-text-field--with-trailing-icon';
     }
+    
+    if ('disabled' in props && props.disabled) {
+      className += ' mdc-text-field--disabled';
+    }
 
     if ('value' in props && this.state.showFloatingLabel) {
       className = [className, 'mdc-text-field--upgraded'].join(' ');


prateekbh:
It is already supported.
Check this: https://material.preactjs.com/component/textfield

GusRuss89:
Ah I see. The problem for me was that the `_mdcProps` are expected to be boolean, whereas regular p/react boolean attributes are just evaluated as truthy or falsey.

For this reason I was getting the `disabled` attribute on the input field, but not getting the class on the container.

Is this by design? It seems to me like it should follow the standard for how boolean attributes are handled elsewhere. This also means my code has to be this:

`disabled={props.matches.hasOwnProperty('email')}`

instead of this...

`disabled={props.matches.email}`

prateekbh:
can you create a repro for me on code sandbox?
By design it should put disabled on TextField and add the class

GusRuss89:
Sure, here you go - https://codesandbox.io/s/xj6x3xmqvz

prateekbh:
Ah got it!
I wouldn't recommend doing it for one prop for one component, I guess a better approach would be to change it here:
https://github.com/prateekbh/preact-material-components/blob/master/MaterialComponent.js#L41

and we can discuss on that from there on

CC: @YePpHa @filoozom @mellogarrett @developit

GusRuss89:
Yes agree on that, I didn't realise when I wrote the pull request that that's how it worked.

PR #791, opened 2018-05-09
select bug fix

prateekbh:
undefined issue fix


Select/Select.jsx
```diff
@@ -39,9 +39,11 @@ class Select extends MaterialComponent {
 
     const selectedIndex = this.MDComponent.selectedIndex;
     if (selectedIndex === 0) {
-      this._labelRef.classList.remove('mdc-floating-label--float-above');
+      this._labelRef &&
```
YePpHa: To be honest it would probably be more readable if you used an if statement.

PR #830, opened 2018-06-17
Default id for TextFieldInput

pavex:
If showDiv is false, id is not available for TextFieldInput when id not defined in props. Generate id for both.


artipost:
Failed pictures (2): [failed-pictures.zip](https://gist.githubusercontent.com/artipost/fb07f367dc12af8fe931e5c9b423749f/raw/d32632a33ef57194858e82ed17b8c7fbe8ba03e0/failed-pictures.zip)

##
:postbox: Posted by [artipost.io](https://artipost.io), through Travis build [393257804](https://travis-ci.org/prateekbh/preact-material-components/builds/393257804), for commit [df80019](https://github.com/prateekbh/preact-material-components/pull/830/commits/df8001908d7ab04b70595cabe2e872523df6ab43).

pavex:
@prateekbh Hi, Yes, TextFieldInput needs id when label is set.
better as follows (on line 172):
```
     if (!!props.label && !props.id) {...
```
...but, I do not know what the code with showFloatingLabel is. I think it is totally useless. You could use only one TextFieldInput, with or without a container, depending on whether helperText is available.

prateekbh:
Makes sense.
172 can indeed be changed to your latest suggestion

PR #922, opened 2018-08-08
Added babel runtime as runtime dependecy

cromefire:
As babel is imported at runtime it needs to be present at runtime.


artipost:
Failed pictures (27): [failed-pictures.zip](https://gist.githubusercontent.com/artipost/99527eb8b70b03bebc950f31e762f9dc/raw/2f1ee192eb583278067c9c3dc7544ef2561fc0a3/failed-pictures.zip)

##
:postbox: Posted by [artipost.io](https://artipost.io), through Travis build [413509784](https://travis-ci.org/prateekbh/preact-material-components/builds/413509784), for commit [ab388ae](https://github.com/prateekbh/preact-material-components/pull/922/commits/ab388ae8944b188ae71b12df960bba9f2d844a2b).

prateekbh:
what happens to people using babel 6?

cromefire:
Our code uses babel 7, their code uses babel 6 and they don't save that much space It's simply same as before runtime. It's just a bit better overall and even better if they also use babel 7 and runtime.

cromefire:
For 6 it's `babel-runtime`, for 7 it's `@babel/runtime` and in addition `npm`/`yarn` should handle it appropriately

cromefire:
No that's unnecessary because our code only uses Babel 7, what our code and their code uses is totally separated.

cromefire:
If they import `babel-runtime` (aka. babel 6 + runtime), fine.
If they import none of them, fine.
If they import `@babel/runtime` (aka. babel 7 + runtime), it gets only shipped once, resulting in lower bundle size, still fine.

prateekbh:
so what happens if they use babel 6 and try to use our components?
they get error?

cromefire:
No the final bundle will include both, `babel-runtime` (if they use runtime) and `@babel/runtime`

cromefire:
They just don't get bundles as small as babel 7 + runtime (same for babel 6 and 7 without runtime)

cromefire:
As this is a serious bug, I think it should be released relatively soon

Issue #6, opened 2017-03-17
Check how these perform with server side rendering


prateekbh:
Most of initialization code is written in componentDidMount, verify how the noJS version of these things work.
or even how does the transition from server rendered to client rendering looks like

cromefire:
I will look at this shortly

prateekbh:
yasss thank you for all the love ❤️

cromefire:
Not doing all maybe in a few days, on some you need to decide

cromefire:
```
Unhandled promise rejection ReferenceError: MutationObserver is not defined
    at Object.registerValidationAttributeChangeHandler (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:7074:26)
    at MDCTextFieldFoundation.init (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:5442:48)
    at MDCTextField.MDCComponent (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:379:22)
    at new MDCTextField (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:6884:249)
    at TextFieldInput.eval (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:13048:42)
    at renderComponent (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:15345:37)
    at setComponentProps (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:15204:4)
    at buildComponentFromVNode (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:15370:3)
    at idiff (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:14925:10)
    at innerDiffNode (eval at <anonymous> (/path/to/project/node_modules/prerender-loader/src/index.js:192:21), <anonymous>:15044:12)
```
_(prerender-loader)_

Challenge accepted

cromefire:
Ref #929

cromefire:
Fixed by #982

PR #929, opened 2018-08-16
Adding support for prerendering

cromefire:
- [ ] ~~prerender detection > custom hook~~
- [ ] ~~exclude parts that prevent prerendering~~
- [x] Fix some bugs to make the prerendered version look different

### DIFF
--- a/docs/src/routes/home/index.js
+++ b/docs/src/routes/home/index.js
@@ -29,15 +29,19 @@ export default class HomePage extends Component {
             app, without worrying about the extra code of components which you
             are not using.
           </div>
-          <br />
-          <Typography headline4>How to use</Typography>
-          <br />
-          <br />
-          <Typography headline5>There are three distributions:</Typography>
+          <Typography headline4 tag={'div'}>
+            How to use
+          </Typography>
+          <Typography headline5 tag={'div'}>
+            There are three distributions:
+          </Typography>
           <ul>
             <li>
-              <Typography headline6>Default Build</Typography>
-              <div>Location:</div> Package Root /{' '}
+              <Typography headline6 tag={'div'}>
+                Default Build
+              </Typography>
+              <div>Location:</div>
+              Package Root /{' '}
               <span className="code-snippet">
                 {'preact-material-components/<Component>'}
               </span>
@@ -45,7 +49,7 @@ export default class HomePage extends Component {
               <ul>
                 <li>
                   Libraries: ✖ (Users should decide about transpilation
-                  themselves)
+                  themselves, but it is possible)
                 </li>
                 <li>Applications: ✔</li>
               </ul>
@@ -57,10 +61,11 @@ export default class HomePage extends Component {
             </li>
             <br />
             <li>
-              <Typography headline6>ES Module Build</Typography>
-              <div>
-                Location:
-              </div> <span className="code-snippet">esm</span> directory /{' '}
+              <Typography headline6 tag={'div'}>
+                ES Module Build
+              </Typography>
+              <div>Location:</div>
+              <span className="code-snippet">esm</span> directory /{' '}
               <span className="code-snippet">
                 {'preact-material-components/esm/<Component>'}
               </span>
@@ -71,22 +76,25 @@ export default class HomePage extends Component {
               </ul>
               <div>Support:</div>
               <div>
-                This supports most recent browsers, JavaScript and TypeScript
+                This supports most recent browsers, JavaScript and TypeScript,
+                if you want to transpile it yourself use this version
               </div>
             </li>
             <br />
             <li>
-              <Typography headline6>TypeScript Source</Typography>
-              <div>Location:</div> <span className="code-snippet">ts</span>{' '}
-              directory /{' '}
+              <Typography headline6 tag={'div'}>
+                TypeScript Source
+              </Typography>
+              <div>Location:</div>
+              <span className="code-snippet">ts</span> directory /{' '}
               <span className="code-snippet">
                 {'preact-material-components/ts/<Component>'}
               </span>
               <div>Who?</div>
               <ul>
                 <li>
-                  Libraries: (✔) (Not recommended, keep in mind that your
-                  library will only support TypeScript)
+                  Libraries: ✖ (Not recommended, keep in mind that your library
+                  will only support TypeScript)
                 </li>
                 <li>Applications: ✔</li>
               </ul>
@@ -98,19 +106,19 @@ export default class HomePage extends Component {
             </li>
           </ul>
           <br />
-          <Typography headline5>Installation and Usage:</Typography>
+          <Typography headline5 tag={'div'}>
+            Installation and Usage:
+          </Typography>
           <div>
             All the components of{' '}
             <span className="grey">preact-material-components</span> are built
             to work independently. So there are couple of ways to use them.
           </div>
+          <div class={'code-snippet'}>npm i -D preact-material-components</div>
           <div>
-            <CodeBlock>
-              <code class="lang-js">npm i -D preact-material-components</code>
-            </CodeBlock>
-          </div>
-          <div>
-            <Typography headline6>Using the JS of the component</Typography>
+            <Typography headline6 tag={'div'}>
+              Using the JS of the component
+            </Typography>
             <div>
               You can also use the components individually, as none of the tree
               shaking currently removes unused classes.
@@ -128,7 +136,9 @@ export default class HomePage extends Component {
             </div>
           </div>
           <div>
-            <Typography headline6>Using the CSS of the component</Typography>
+            <Typography headline6 tag={'div'}>
+              Using the CSS of the component
+            </Typography>
             <div>
               If you are using only a couple of components from the entire
               package, try importing individual CSS, for the same reason: TO
--- a/docs/src/routes/prerendering/index.js
+++ b/docs/src/routes/prerendering/index.js
@@ -6,6 +6,8 @@ import Typography from '../../../../Typography';
 
 // Components
 import CodeBlock from '../../components/code-block';
+import renderToStringSample from './render-to-string.txt';
+import renderToStringPLSample from './render-to-string-prerender-loader.txt';
 
 // Samples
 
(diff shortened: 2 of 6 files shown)


cromefire:
There are problems with `MutationObserver`s. Maybe we can polyfill them?

prateekbh:
Quick question, who is using MutationObserver???

I thought it's a debugging thing

cromefire:
MWC, it notifies you, if you alter a DOM element.

cromefire:
I just excluded all the JS Components if the user enables "prerender mode", by simply calling a static method (the only thing that seems to be altered seems to be TextField, as it needs the upgraded class, if the JS Component is enabled (now dynamically added)).

cromefire:
But as I think about it, there might be a bug with that.

cromefire:
I don't know which bug you mean.

For the mutation observer I think it might be all, but now I'm excluding Javascript from the components, so it's fine. 

The other one is me looking, if `this.MDComonent` is defined in `render()`, but `render()` is executed before `componentDidMount()`.

artipost:
Failed pictures (1): [failed-pictures.zip](https://gist.githubusercontent.com/artipost/531dc5b0c5800dfe3eaccc77fe05a108/raw/e075f62c3d659590d43e879157f6ad7d86421ebc/failed-pictures.zip)

##
:postbox: Posted by [artipost.io](https://artipost.io), through Travis build [423326373](https://travis-ci.org/prateekbh/preact-material-components/builds/423326373), for commit [e4b0ef8](https://github.com/prateekbh/preact-material-components/pull/929/commits/e4b0ef8291d93c611074c6d8f47b988b9f9858b1).

artipost:
Failed pictures (1): [failed-pictures.zip](https://gist.githubusercontent.com/artipost/183a0a8e31fa4d31a14782456033919a/raw/99a4e671c87cec6c7eae409a07242df86d019695/failed-pictures.zip)

##
:postbox: Posted by [artipost.io](https://artipost.io), through Travis build [423326713](https://travis-ci.org/prateekbh/preact-material-components/builds/423326713), for commit [62251b4](https://github.com/prateekbh/preact-material-components/pull/929/commits/62251b4fe975e81834c2424781bbfd9ca7885f4c).

artipost:
Failed pictures (1): [failed-pictures.zip](https://gist.githubusercontent.com/artipost/00de50e3a82d9e385bf62f06709589a1/raw/ae8681581f433a85941f74bdf3911165e9727402/failed-pictures.zip)

##
:postbox: Posted by [artipost.io](https://artipost.io), through Travis build [423331770](https://travis-ci.org/prateekbh/preact-material-components/builds/423331770), for commit [9ac8f38](https://github.com/prateekbh/preact-material-components/pull/929/commits/9ac8f38370fddd5267fb09f04baa169536a53e64).

artipost:
Failed pictures (1): [failed-pictures.zip](https://gist.githubusercontent.com/artipost/088474388c3b78cf4ae1eba88751487b/raw/259b2f28f9d8fc389ef5ef5a6077a22727ece901/failed-pictures.zip)

##
:postbox: Posted by [artipost.io](https://artipost.io), through Travis build [423336630](https://travis-ci.org/prateekbh/preact-material-components/builds/423336630), for commit [19bcd63](https://github.com/prateekbh/preact-material-components/pull/929/commits/19bcd638504a079a6963606aed8db1e6aa7bd5ae).

prateekbh:
I meant like server side rendering of which component gives MutationObserver error?

ts/Checkbox/index.tsx
```diff
@@ -31,7 +23,7 @@ export class Checkbox extends MaterialComponent<
 
   public componentDidMount() {
     super.componentDidMount();
-    if (this.control) {
+    if (this.control && !MaterialComponent.isPrerendering) {
```
prateekbh: cDM should never run in a server side code by preact's design. Why this check is even required?

cromefire: I depends on the technique you use, many frameworks just render the page either in the browser or in jsdom (Which doesn't have Mutation observer for example).

cromefire: I agree that there are specific tools like `preact-cli` which don't do that.

cromefire: See https://github.com/GoogleChromeLabs/prerender-loader#string-prerendering for example

cromefire:
> I meant like server side rendering of which component gives MutationObserver error?

Sorry can reproduce that easily any more, as I would have to revert a good part of my app. I would think that's something used in many components.

ts/Checkbox/index.tsx
cromefire: > headless outdated browser

`jsdom` is no browser, it runs in nodejs and it's used fairly often, if you don't want a headless browser. Headless browser would just work fine.

> react/preact has a first class server side rendering

Haven't seen server side rendering for preact yet

prateekbh: Preact has `render to string` just like react, why would you need anything else?

prateekbh:
>Okay after searching, I found out, that only the TextField uses MutationObservers

With `render-to-string` mdc-web` will never be initialized. So that shouldn't be a problem.

ts/Checkbox/index.tsx
cromefire: Oh I couldn't find this because it's in an separate library

cromefire:
Yeah but render to string wasn't in the main docs, so I didn't find it

cromefire:
Okay tested it, that works I would prefer to write some docs about that anyway so other people can reproduce that easily

cromefire:
So the problem right now is, that this is, how it should look:
desired
And this is how it looks:
current
As you can see preact adds another instance of the UI instead of augmenting the prerendered one
(prerendered on the left and generated on the right).

prateekbh:
IIUC you are saying that it is generating duplicate elements?

prateekbh:
>Yeah but render to string wasn't in the main docs, so I didn't find it

Sorry this might've been my ignorance I thought it's a well known pattern in p/react world.
We can add it to docs

cromefire:
Yeah I knew it for React but couldn't find it in the `preact.d.ts` file

> IIUC you are saying that it is generating duplicate elements?

Yes although they have the same `id`

prateekbh:
> Yes although they have the same id.

I've faced this earlier, it's a thing in preact not pmc.
Can you close this and show me the code u use to do SSR. I might be able to help

cromefire:
> I've faced this earlier, it's a thing in preact not pmc.

Yes I know, but I thought you know preact a lot better than I do

> Can you close this

This still fixes some things for prerendering to make the difference between the prerendered and the actual smaller, plus it will add some docs to avoid that others have that problems.

> show me the code u use to do SSR

``` typescript tsx
renderToString(<UI emitter={emitter} />);
```
The string is simply put into body.
Client side:
``` typescript tsx
render(<UI emitter={emitter} />, document.body);
```

cromefire:
Okay after some trial and error I found out that I have to render to
```typescript
document.getElementById("UI")
```
if I have prerendered content.

Now it's not duplicate, but also doesn't update the prerendered content, so it's like it's not rendering.

prateekbh:
>This still fixes some things for prerendering to make the difference between the prerendered and the actual smaller

Can you list the exact cases this solves?

>The string is simply put into body. \n Client side:

Yep thats where the bug is, here's a sample on how to do it: 

```
render(<UI emitter={emitter} />, document.body, document.body.firstElementChild);
```

cromefire:
> Yep thats where the bug is, here's a sample on how to do it:

Yes, found that out yesterday using the `hot` example, somehow I think I forgot to send the comment. 

> Can you list the exact cases this solves?

As long, as the Javascript component is not attached the `TextField` it uses placeholder instead of a label above the `TextField`, plus various code cleanup things. In addition to that I will start adding docs for it, now that it works.

cromefire:
> `document.body.firstElementChild` you have to be careful that's not 100% universal, I for example have some thing in my template that's not managed by preact, but an handwritten error page.

cromefire:
[prerendered-in-action.webm.zip](https://github.com/prateekbh/preact-material-components/files/2345532/prerendered-in-action.webm.zip)

There you can see the difference between the prerendered and online version of the page, the difference between the placeholder and the floating label is indistinguishable as long as it is not filled (else the floating label will pop up when the JavaScript activates).

cromefire:
Fixes #6

prateekbh:
Ok I see what u meant there. But I see this as a bug from mdc and not us. Also the difference is kinda important as it would be unusable without jS aka "progressive enhancement". But it could be definitely be better.

I am not super leaned for doing any thing specialized when the out of the box solution works perfectly for every component except one use case.

Entire material.preactjs.com is pre-rendered and is fine

#6 was just to check if there is any perf regression for these components on server side rendering

cromefire:
Yes, but I think it's really quite good how it is now, because before there was that one none mdc text field issue I fixed.

> Entire material.preactjs.com is pre-rendered and is fine

screenshot from 2018-09-04 16-52-28
That a thing solved by this PR because that's something that isn't from mdc, but I agree the rest looks pretty okay.

cromefire:
It's not that bad, but it caused whole UI to rearrange, when the javascript got activated

cromefire:
> I am not super leaned for doing any thing specialized when the out of the box solution works perfectly for every component except one use case.

What do you mean with _anything specialized_

> #6 was just to check if there is any perf regression for these components on server side rendering

I wanted GitHub to close the issue when this is merged, but I think that's not possible from comments and performance wise, I couldn't see any problem.

prateekbh:
+1 it should be fixed, but I still believe it should be from mdc team and not us. It's a lot of stuff for fixing a very nice problem

cromefire:
> but I still believe it should be from mdc team and not us

mdc cannot fix that, they never said to put that label there, it's our problem

cromefire:
It's added outside of the `mdc-text-field`

cromefire:
What they say is like when you only use css TextField you cannot use floating label, use some thing different.

So before we used a label outside of the TextField (not the floating label, just a plain `<label>`) and now we use placeholder (also plain html) which looks more similar to the floating label.

prateekbh:
>mdc cannot fix that, they never said to put that label there, it's our problem 
so we add this when user adds a [prop](https://github.com/prateekbh/preact-material-components/blob/master/ts/TextField/index.tsx#L163-L166) `label`

If someone wants to add a label, this is how they'd add it even in vanilla js mdc-web. no? If there is a better way of doing it in vanilla js but we regressed i'll be more than happy to fix it.

prateekbh:
>So before we used a label outside of the TextField (not the floating label, just a plain <label>) and now we use placeholder (also plain html) which looks more similar to the floating label.

We can fix this using a state variable in textfield and only for textfield not something which affect everything.

Perhaps, this could be a very small PR.

cromefire:
> If someone wants to add a label, this is how they'd add it even in vanilla js mdc-web. no?

If I'd use plain mdc I would use placeholder, but label is possible.

> Perhaps, this could be a very small PR.

In fact 90% of the changes in this PR are like various improvements I've done (like removing unused functions, etc).

> We can fix this using a state variable in textfield and only for textfield not something which affect everything.

This is how it's done, I just did not rename the state variable back, but the other things got reverted in 
0610866.

artipost:
Failed pictures (2): [failed-pictures.zip](https://gist.githubusercontent.com/artipost/9dae43d7145912aeb74f1930fff0f53c/raw/0ce49a0d2cb4f886d3ef97fc684b17d04bba8946/failed-pictures.zip)

##
:postbox: Posted by [artipost.io](https://artipost.io), through Travis build [424407518](https://travis-ci.org/prateekbh/preact-material-components/builds/424407518), for commit [12b13f7](https://github.com/prateekbh/preact-material-components/pull/929/commits/12b13f728de1798fe8ccd5ee77608770d95113b5).

cromefire:
So to be clear this does not affect everything anymore, forget anything from the beginning of this PR `render-to-string` solved it. It's now just the one minor `TextField` fix.

prateekbh:
just a thought to make this easier, lets open another PR just to fix textfield.
about docs on pre-rendering, I am not sure if we need this. The topic is well beyond our package, a better thing might be a medium blog instead of a adding a page on this link.

PR #930, opened 2018-08-17
TextField component was missing ITextFieldProps

r1cs1:


cromefire:
I will check that again but if I'm not wrong `OmitAttrs` _should_  include `ITextFieldProps`.

cromefire:
@prateekbh merge it

cromefire:
Should be quite stable, works well in my application since a while now.

Issue #702, opened 2018-04-02
support for accessing value on TextField items


uktris:
Hi,

I've been testing out preact for a project and working with the material component library for the look and feel. So far it's all been going great but I've noted what I think is an omission but it would be good if someone could verify...

I occasionally use refs to deal with components that really don't need state. With the TextField component this allows access to the main object and I can get to the input field value with the nested this.refname.MDComponent.input_.value but it feels a little wrong to do so.

I don't know which is the standardized approach but I see a few options such as a) getValue() function on the TextField object b) support inputRef to get a reference to the underlying text field c) propagate the value up to the TextField in "value".

So somewhere between a bug and a feature request I guess :)

Tristan.

cromefire:
Yes looking into doing this

cromefire:
For the value you can actually use
``` typescript
this.refname.MDComponent.value
```

cromefire:
But that only work's for `value`

PR #932, opened 2018-08-22
Proper refs for TextField and ref typing

cromefire:
- [x] `TextField` refs can properly access `TextFieldInput`
- [x] All Elements that are wrappers around an `<input>` element allow access to a reference of that input element
- [ ] Type all refs

A ref of `TextFieldInput` can be accessed via
``` typescript
text_field_ref.tfInput
```

Input refs will be accessible via
``` typescript
ref.input
```

### DIFF
--- a/ts/Base/InputComponent.ts
+++ b/ts/Base/InputComponent.ts
@@ -1,9 +1,11 @@
 import MDCComponent from '@material/base/component';
+import MDCFoundation from '@material/base/foundation';
 import autobind from 'autobind-decorator';
 import MaterialComponent from './MaterialComponent';
 
 export abstract class InputComponent<
-  M extends MDCComponent<any, any> & {value: string},
+  M extends MDCComponent<any, F>,
+  F extends MDCFoundation<any> & {getValue(): string},
   PropType,
   StateType
 > extends MaterialComponent<PropType, StateType> {
@@ -20,8 +22,8 @@ export abstract class InputComponent<
   }
 
   public get value(): string | undefined {
-    if (this.MDComponent && this.MDComponent.value) {
-      return this.MDComponent.value;
+    if (this.MDComponent) {
+      return this.MDComponent.getDefaultFoundation().getValue();
     } else if (this.inputRef) {
       return this.inputRef.value;
     }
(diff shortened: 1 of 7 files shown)


prateekbh:
What's are we trying to achieve with this?

prateekbh:
> All Elements that are wrappers around an <input> element allow access to a reference of that input element

I would discourage this. The underlying Dom is not for us to own. A `ref` to this would make a breaking change for us even though mdc APIs remain same

cromefire:
> A `ref` to this would make a breaking change for us even though mdc APIs remain same

Where do you see a breaking change?

> The underlying Dom is not for us to own

Is't pure preact

> What's are we trying to achieve with this?

Currently the only way to access the input element (see #702, e.g. for calling DOM methods on it) is either to use `MDComponent._input` (what's definitely discouraged) or traversing the DOM your self (using `document.getElementById()` etc.; that's also discouraged, because as you said, the DOM is not owned by us).

cromefire:
Example for a DOM API not supported by MDCComponents: `reportValidity()`

prateekbh:
> Where do you see a breaking change?

E.g. Select was never implemented using the <select> component until recently, if someday they go back it won't be a breaking change for them as the APIs will be same but it will be for us as we won't have a Dom to give ref to.

Therefore I would suggest to own only mdc properties and not have any additional logic on us.

Plus a Dom ref to top level always exposes inner Dom to work with.

cromefire:
Traversing through the inner Dom is really manual, there are use cases that **need** that Dom ref, so the alternative is to tell the users to use some hacks to that will also have that breaking changes and are not designed to be used like this. That means that this would be a breaking change from MWC which we cannot influence any ways.
Maybe would do better to implement that and then mark it as "Only use it, if you really need"

prateekbh:
> Maybe would do better to implement that and then mark it as "Only use it, if you really need"

I've had some discussion to add some kind of "experimental"  features to this package always. Just never had enough time to do so.

Do u have a good API design to do so?

cromefire:
> Do u have a good API design to do so?

The only thing I can think of right now is annotating it in the docs and annotating it in the source.

cromefire:
Dopping this in favor of the refactor for `2.0`

cromefire:
Please leave the branch for now

Issue #970, opened 2018-08-28
Programatically control active tab for tabbar


r1cs1:
Hi guys,

I'm trying to programatically control active tab for tabbar, however neither activeTabIndex on TabBar, nor active prop on Tab is working.

Am I missing something or that's expected?

Thanks,
Richard

cromefire:
Can you share how you try to do this? I think I know what's going wrong and if so a fix is on the way.

replaysMike:
Oh man I just came here to ask that exact question as well. I'll show you what I did, but it's an awful solution in my opinion but the docs here are pretty scarce in detail.

```
handleTabChange = (e) => {
	// todo: this feels old school, do Tabs component handle content selection?
	const targetName=e.toElement.getAttribute('tab');
	const panelsContainer = document.getElementById('panels');
	var panels = panelsContainer.getElementsByTagName('*');
	for(var i=0,len=panels.length; i<len; i++){
		panels[i].style = 'display: none;'
	const target = document.getElementById(targetName);
	target.style = 'display: block';
	};
}

render(props, state) {
	return (
		<div>
			<Tabs onClick={this.handleTabChange}>
				<Tabs.Tab tab="panel-1" active={true}><Icon>favorite</Icon>Texas Holdem</Tabs.Tab>
				<Tabs.Tab tab="panel-2">Blackjack</Tabs.Tab>
				<Tabs.Tab tab="panel-3">Guts</Tabs.Tab>
			</Tabs>
			<section id="panels">
				<p id="panel-1">Texas Holdem Content</p>
				<p id="panel-2" hidden={true}>Blackjack Content</p>
				<p id="panel-3" hidden={true}>Guts Content</p>
			</section>
		</div>
	);
}
```

I hope someone else can chime in and indicate the proper way to do things, I've spent a few hours looking at the examples and source with no answer.

replaysMike:
In addition to that, for me I get this when trying to use the TabBar (since Tabs is marked as deprecated):

`Module not found: Error: Can't resolve 'preact-material-components/TabBar'`
¯\_(ツ)_/¯

r1cs1:
Here's a dummy example, because my code too complex to paste here. I hope I didn't mess it up .

Also, I'm using 1.5.1-beta1 because that's the only one that compiles with little mods with my typescript setup

```
handleClick = event => {
  this.setState({ active: +event.target.dataset.index });
}

render() {
  return (
    <TabBar>
[...]

PR #971, opened 2018-08-29
Improved TypeScript typing for soft merge

cromefire:
See https://github.com/prateekbh/preact-material-components/issues/970#issuecomment-416740564

### DIFF
--- a/ts/Base/MaterialComponent.ts
+++ b/ts/Base/MaterialComponent.ts
@@ -2,7 +2,7 @@ import MDCComponent from '@material/base/component';
 import {MDCRipple} from '@material/ripple';
 import autobind from 'autobind-decorator';
 import {Component, VNode} from 'preact';
-import {OmitAttrs} from './types';
+import {SoftMerge} from './types';
 
 export interface IMaterialComponentOwnProps {
   ripple?: boolean;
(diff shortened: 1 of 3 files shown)


cromefire:
Did for that off `mdc-notify` by accident, so has to be merged back there, but shouldn't be that bad because this is not a really big change.

prateekbh:
After this are we good for a stable?

cromefire:
If we can quickly resolve that `MDCTab` thing in the other PR and you don't have anything to complain about I think it is.

cromefire:
> If we can quickly resolve that `MDCTab` thing

Or at least agree on looking at this later again

cromefire:
So if anyone wants to look over this (for example @YePpHa as TS expert), else I'll merge this

prateekbh:
MdcTab thing is not prop friendly as of now... i posted it there and an example.

cromefire:
I just found it in my code completions and I'm not sure, if it should be attached to the tab.
It seems like it should be the case ([see here] (https://material.io/develop/web/components/tabs/tab/))

cromefire:
You know, let me open a PR for this and then we can see

cromefire:
For now as nobody has a complaint, I'll merge this

PR #979, opened 2018-09-05
Replaced autobind-decorator with bind-decorator

cromefire:
For smaller bundles

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3523,11 +3523,6 @@
       "integrity": "sha1-ri1acpR38onWDdf5amMUoi3Wwio=",
       "dev": true
     },
-    "autobind-decorator": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/autobind-decorator/-/autobind-decorator-2.1.0.tgz",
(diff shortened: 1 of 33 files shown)


prateekbh:
This one throws better error messages, already we know there's an ear now.

prateekbh:
Can you please take a look at the errors

prateekbh:
Also why does `materialDom` even needs this???
I thought only event listeners will need this

cromefire:
> Can you please take a look at the errors

It's a OSX/fsevents error, it seems that they have a broken/at least not prebuild version

prateekbh:
Nups, I am taking about this
https://travis-ci.org/prateekbh/preact-material-components/jobs/424755513#L456

cromefire:
> Also why does materialDom even needs this???
I thought only event listeners will need this

I could swear I had problems with that, but now it seems to work

cromefire:
> Nups, I am taking about this
https://travis-ci.org/prateekbh/preact-material-components/jobs/424755513#L456

I think you can't override methods annotated with`@bind`.

PR #980, opened 2018-09-05
Optimized named exports

### LINKED ISSUE #134: Component/index.js commited
Hey, I don't know why but, when I build the project all the index.js files are different from those in the repo, even if I didn't touch the matching jsx file.

The two main differences are : 
1. The _extends function which is a generated as a one liner on my machine.
2. The call to _extends which is not indented the same way.

As these files are in the git repo I can't just commit my changes without taking extra care of not committing the compiled version of the jsxs.

Do we really need to publish them to git ?

cromefire:
Fixed #134


ts/IconButton/index.tsx
```diff
@@ -4,6 +4,8 @@ import {h} from 'preact';
 import MaterialComponent from '../Base/MaterialComponent';
 import Icon from '../Icon';
 
+export {Icon as IconButtonIcon} from '../Icon';
```
YePpHa: What's the purpose of this? Does it really make sense to export it from IconButton as well as Icon?

cromefire: It's not totally needed but it's nice to have and doesn't add any overhead

cromefire:
@prateekbh you can merge this

cromefire:
> Let's not advertise this till 2.0 though

Agree

ts/Dialog/index.tsx
```diff
@@ -2,7 +2,7 @@ import {MDCDialog} from '@material/dialog/';
 import {bind} from 'bind-decorator';
 import {h} from 'preact';
 import MaterialComponent from '../Base/MaterialComponent';
-import Button from '../Button';
+import {Button} from '../Button';
```
prateekbh: Shouldn't button be a default export?

cromefire: It has a default export but I don't know, if things like webpack can treeshake when you use the default export

Issue #999, opened 2018-09-17
Add mdc-tab/scroller packages


prateekbh:
Mdc-tab-bar's styles are segregated into separate packages, we need to add these packages for their style.css or just copy these styles into mdc-tab-bar's.

Later might be easier

prateekbh:
@cromefire can u take this up?

cromefire:
So the style from scrolled just have to end up in the Tab directory?

prateekbh:
Well I thought about this and adding the packages would be better as our script will copy the case and scss. No need to implement the js

cromefire:
> No need to implement the js

Didn't implement any js See here #1002

PR #994, opened 2018-09-12
fixing changes in 39

prateekbh:
- fixing dom for drawers
- fixing doc's site
- fix icon button
- fixing drawers docs

PR #1002, opened 2018-09-17
Copy multiple css packages to one component directory

cromefire:
The structure changes to:
- `css`: `preact-material-components/<Component>/css/<package>.css`
- `scss`: `preact-material-components/<Component>/scss/<package>/**/*.scss`

Ref #999


prateekbh:
sorry i was late on this.
This would be a huge breaking change and thats the reason i wanted to fallback to just adding the components to our repo instead of copy multiple css to one folder.

But even if we want to do this, the structure should not change. Thats an unnecessary change

prateekbh:
for reference i added one similar pckg in https://github.com/prateekbh/preact-material-components/pull/994

prateekbh:
#1007 is fixing this for now, lets compile eeverything scss starting 2.0

PR #1042, opened 2018-09-28
fixing drawer's MDComponent

### LINKED ISSUE #1010: "Can't have a focus-trap without a focusable element" in Drawer
I just upgraded my app to 1.5.3 from the master branch and the only thing that I haven't gotten to work again is the Drawer component, which raises the following exception:
`index.js?79b7:169 Uncaught Error: You can't have a focus-trap without at least one focusable element`

focus-trap

This also happens when I copy the updated example here:
https://github.com/prateekbh/preact-material-components/blob/master/docs/src/routes/drawer/sample.txt

prateekbh:
- fixes #1010 (the problem of being MDComponent as protected)


prateekbh:
I didn't get you

cromefire:
Fix it there:

https://github.com/prateekbh/preact-material-components/blob/50fc73f5e5ccedb99a5e13d863cfb6307e563843/ts/Base/MaterialComponent.ts#L58

cromefire:
> It's okay as a hacky way until 2.0, but please change it to public in MDComponent

Sorry I meant `MaterialComponent`

prateekbh:
@cromefire done, but until 2.0 lets keep shifting from protected to public in all components one by one

prateekbh:
@cromefire we are all set to release 1.5.3 right?

cromefire:
Let me check

cromefire:
See https://github.com/prateekbh/preact-material-components/milestone/3

cromefire:
2 left, 1 of them ready for merging, 1 of them small

PR #1045, opened 2018-09-29
Remove unneeded argument to render for TextField

cromefire:
Removed unnecessary state argument for TextField


cromefire:
Just a quick fix

PR #1047, opened 2018-09-30
Fixed value update on slider

### LINKED ISSUE #1009: Slider not re-rendered when the value prop changes
As the title says, I'm using the Slider component and noticed that the thumb stays at its old position when the value prop changes.  I think what's missing is only an implementation of componentWillReceiveProps which checks if the value has changed and calls setValue if necessary?

cromefire:
Fixes #1009


cromefire:
The last PR after this, `1.5.3` is done and `2.0` is the goal

PR #1049, opened 2018-10-01
Fixed mdcNotifyProps super calls

cromefire:
Some super calls where forgotten to implement, so `mdcNotifyProps` didn't work.

Also they are now in `componentWillReceiveProps` instead of `componentWillUpdate`, I don't know the difference, but it seems right


cromefire:
@prateekbh Could you please confirm this bug-fix?

ts/LinearProgress/index.tsx
```diff
@@ -31,8 +31,8 @@ export class LinearProgress extends MaterialComponent<
     this.afterComponentDidMount();
   }
 
-  public componentWillUpdate(nextProps: ILinearProgressProps) {
-    super.componentWillUpdate(nextProps);
+  public componentWillReceiveProps(nextProps: ILinearProgressProps) {
```
prateekbh: add a todo to to switch away from `cWRP` as it is now deprecated

cromefire:
Should we use componentWillUpdate instead?

ts/LinearProgress/index.tsx
cromefire: 2.0

cromefire:
Added an issue to remove all calls until `2.0`

prateekbh:
All will* callbacks are deprecated

cromefire:
And how are we supposed to handle that?

cromefire:
So should we do everything in componentDid*?

prateekbh:
getsnapshotbeforeupdate/ componentDidUpdate/useEffect(because hooks)

cromefire:
Okay, I will look those up

PR #1190, opened 2018-12-06
Switch: Use mdcNotifyProps

### LINKED ISSUE #1189: Switch not responding to props change
When the props (`selected `, `disabled`) changes after a Switch component is initiated , it does not change its shape accordingly, thus I cannot create a controlled Switch.
The props only affects the initial render of a Switch.
A quick example to reproduce the bug: 
https://codesandbox.io/s/jvkx6zoj75

This might be caused by the upstream as stated here:
https://github.com/material-components/material-components-web/tree/master/packages/mdc-switch#initially-disabled-switch
Maybe we should use `MDCSwitchFoundation.setDisabled` and `MDCSwitchFoundation.setChecked`?

eligao:
I've removed all updates of `MDComponent.value`, and used `mdcNotifyProps` instead, to make the switch respond to the change of `props.disabled` and `props.checked`.
I'm not sure why `MDComponent.value` is updated here, it is neither existing here in `@material/switch` nor documented.
https://github.com/material-components/material-components-web/blob/master/packages/mdc-switch/index.js

Correct me if I'm wrong.

fixes #1189

### DIFF
--- a/ts/Switch/index.tsx
+++ b/ts/Switch/index.tsx
@@ -12,20 +12,12 @@ export class Switch extends MaterialComponent<ISwitchProps, ISwitchState> {
   public MDComponent?: MDCSwitch;
   protected componentName = 'switch';
   protected mdcProps = ['disabled', 'checked'];
+  protected mdcNotifyProps = ['disabled', 'checked'];
 
   public componentDidMount() {
     super.componentDidMount();
     if (this.control) {
       this.MDComponent = new MDCSwitch(this.control);
-      if (this.props.value) {
-        this.MDComponent.value = this.props.value;
-      }
-    }
-  }
-
-  public componentWillReceiveProps(nextProps: ISwitchProps) {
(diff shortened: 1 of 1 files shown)


ts/Switch/index.tsx
```diff
@@ -12,20 +12,12 @@ export class Switch extends MaterialComponent<ISwitchProps, ISwitchState> {
...
       this.MDComponent = new MDCSwitch(this.control);
-      if (this.props.value) {
-        this.MDComponent.value = this.props.value;
-      }
-    }
-  }
-
-  public componentWillReceiveProps(nextProps: ISwitchProps) {
-    if (nextProps.value && this.props.value !== nextProps.value) {
-      this.MDComponent.value = nextProps.value;
```
cromefire: Why did you remove this?

eligao: see my edited post above

cromefire:
Thanks for you work and patience

PR #1242, opened 2019-01-25
Fixing access on value

### LINKED ISSUE #1241: Label and fullwidthTextField
I am getting below warning in console and this seems to be messing up the rendering of the component. Is this a bug or a know restriction ? I have an address field which uses the full width property. Can you please guide me here ? 

Passing a "label" prop is not supported when using a "fullwidth" text field.

cromefire:
Fixes #1241


cromefire:
@prateekbh merge pls

PR #1278, opened 2019-02-03
Better building & testing

cromefire:
Improving build:
- [x] es2015 and esnext build


cromefire:
@prateekbh Check this out:
```
yarn run v1.13.0
$ ts-node -P scripts/tsconfig.json scripts/compile.ts
Collecting packages...
Preparing compilation...
Compiling: base
Compiling: icon
Compiling: button
Compiling: card
Compiling: checkbox
Compiling: chip
Compiling: dialog
Compiling: list
Compiling: drawer
Compiling: elevation
Compiling: fab
Compiling: form-field
Compiling: icon-button
Compiling: image-list
Compiling: layout-grid
Compiling: line-ripple
Compiling: linear-progress
Compiling: menu
Compiling: menu-surface
Compiling: radio
Compiling: select
Compiling: slider
Compiling: snackbar
Compiling: switch
Compiling: tab
Compiling: textfield
Compiling: top-app-bar
Compiling: typography
Compiling [============================] 28/28 100%

Done in 29.13s.
```

`tsickle` integrated

artipost:
Failed pictures (50): [failed-pictures.zip](https://gist.githubusercontent.com/artipost/a8a4705247c231d5972cc90fb08e73a3/raw/27339509920f7cf78d551298ae2fec650051b083/failed-pictures.zip)

##
:postbox: Posted by [artipost.io](https://artipost.io), through Travis build [488117440](https://travis-ci.org/prateekbh/preact-material-components/builds/488117440), for commit [5ef2fad](https://github.com/prateekbh/preact-material-components/pull/1278/commits/5ef2fad1eb1babbac10df6393c385a9c94628e2f).

cromefire:
Well turn out, it didn't actually work...

cromefire:
which leaves the better feedback point

cromefire:
@prateekbh I'd like to not only depend on the screenshot testing, but have like karma unit tests. They also work very well in different browsers (so we can do cross-browser testing) and are far less fragile.

cromefire:
Also I spoke with the `tsickle` guys and they meant it's not possible for a library to be compatible to closure without any big restructure

prateekbh:
Yeah I am not sure Keen on closure, and unit testing is great but let's only restrict it to the parts where we are significant custom code, and not in simpler components like button

cromefire:
Yes I think closure is not worth the investment.

For the unit tests I hope to be able to replace the documentation testing, because it's very fragile has to be fixed every time

prateekbh:
Yes. Also I talked to someone who wants to do better visual regression testing for us. Like really nice on screen diffs n all so we can replace current puppeteer with that may be and unit tests for stuff like drawer etc.

cromefire:
It would really nice if the visual regression testing is cross-browser (or at least Chrome and Firefox) as we could speed up Travis build or test in multiple browsers.

prateekbh:
Yes that's what I have been told

cromefire:
Nice, do you know when it can be added?

prateekbh:
Starting Monday 😊

cromefire:
@prateekbh Do you have any idea, how to fix the compile script, so that it does not exit, before it's finished?

prateekbh:
@cromefire i'd need some more details 
are you referring to this: https://travis-ci.org/prateekbh/preact-material-components/jobs/528169978

^ seems to be completed no?

scripts/compile.ts
```diff
@@ -1,40 +1,56 @@
...
+    });
+    const exit = res ? res.code : process.exitCode || 1;
+    if (exit) {
+      console.error('npx failed');
+      process.exit(1);
+    }
+  }
+
+  await compile('tsconfig.json');
+  await compile('tsconfig.next.json');
```
cromefire: @prateekbh I'm referring to this part. It's run on yarn build, but it's finishing before both functions have run

prateekbh: actually it is completed here: https://travis-ci.org/prateekbh/preact-material-components/jobs/528169978#L238

prateekbh: I assume it has something to do with the fact that one of the task is a promise

cromefire: Yes, the script completes, but if you Insert a `log` after `npx` it's never executed

cromefire: And I'm sure it has todo with the promise, but I didn't find a good way to prevent that and `npx` is kind of the only way, other than a child process (which is not the most portable solution), because the `typescript` compiler API doesn't work with composite builds (at least I haven't found a script/example last time looked)

prateekbh: gotcha, yeah even console.log({res}) is not loggin anything. not sure if its actually waiting for npx to finish

cromefire: I know that at least the 2 npx call is never started. I'm not a node developer and haven't work much with it, so I don't know how to force node to wait. Even polls with `setTimeout` didn't work, so I started wondering if `npx` forces to process to exit somewhere and I have overlooked it? I skimmed through the code and only found it setting the Exit code, but maybe I didn't look close enough

prateekbh: how about using [shelljs](https://www.npmjs.com/package/shelljs) instead for shell commands?

cromefire:
Fixed it

prateekbh:
Ready for review?

cromefire:
Yes

.travis.yml
```diff
@@ -1,6 +1,11 @@
 language: node_js
 node_js:
   - '10'
+addons:
+  apt:
+    packages:
+      # Needed for cypress
```
prateekbh: why?
isnt it already working fine?

packages/button/package.json
```diff
@@ -37,7 +37,8 @@
     "@material/button": "0.44.1",
     "@preact-material-components/base": "^2.0.1-alpha.0",
     "@preact-material-components/icon": "^2.0.1-alpha.0",
-    "bind-decorator": "^1.0.11"
+    "bind-decorator": "^1.0.11",
```
prateekbh: shall we remove this?

packages/card/package.json
```diff
@@ -38,7 +38,8 @@
     "@preact-material-components/base": "^2.0.1-alpha.0",
     "@preact-material-components/button": "^2.0.1-alpha.0",
     "@preact-material-components/icon": "^2.0.1-alpha.0",
-    "bind-decorator": "^1.0.11"
+    "bind-decorator": "^1.0.11",
```
prateekbh: shall we remove this?

packages/checkbox/package.json
```diff
@@ -36,7 +36,8 @@
   "dependencies": {
     "@material/checkbox": "0.44.1",
     "@preact-material-components/base": "^2.0.1-alpha.0",
-    "bind-decorator": "^1.0.11"
+    "bind-decorator": "^1.0.11",
```
prateekbh: and this

.travis.yml
cromefire: It was, it crashed and I followed the docs and now works again (I guess there was an update)

cromefire:
You requested smaller PRs so I wanted to put the `bind-decorator` thing into it's own PR

prateekbh:
Ok all looks good, just waiting for the tests to pass

cromefire:
It will not pass because I downvoted them, because the images are already fixed in other PRs.
But I Reviewed them and it's okay.

cromefire:
Also could you please mark `2.0` as protected branch, so I can't push to it by accident?

prateekbh:
If the tests broke in this PR shouldn't we fix them in the same PR as well?

cromefire:
They didn't break in this PR, but because applitools works global, there are unmerged PRs  changing the global state

PR #1289, opened 2019-02-26
Dialog: change to dialog element

### LINKED ISSUE #1288: Render Dialog in dialog element to be AXE compliant
According to [AXE](https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L279) an alertdialog should be of element type `dialog` or `section`. I'll try to make a PR for this.

kontrollanten:
According to https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L279
an alertdialog should be of element type dialog or section.

Closes #1288

### DIFF
--- a/ts/Dialog/index.tsx
+++ b/ts/Dialog/index.tsx
@@ -132,10 +132,10 @@ export class Dialog extends MaterialComponent<IDialogProps, IDialogState> {
 
   protected materialDom(props) {
     return (
-      <aside role={'alertdialog'} ref={this.setControlRef} {...props}>
+      <dialog role={'alertdialog'} ref={this.setControlRef} {...props}>
         <div className="mdc-dialog__surface">{props.children}</div>
         <div className="mdc-dialog__backdrop" />
-      </aside>
+      </dialog>
     );
   }
 }


cromefire:
I'm not sure that this is the right place, maybe it suited better at [mdc](https://github.com/material-components/material-components-web), they specify the html

kontrollanten:
In their docs they're using `div` element (which is also wrong, but not `aside`), so this shouldn't be aligned with mds? https://github.com/material-components/material-components-web/tree/master/packages/mdc-dialog

cromefire:
Well I guess the dialog doesn't break anything right? The only problem it that you should also create a PR against `2.0` because we already split them off.

prateekbh:
ummm doesn't `dialog` have its own functionality for showing a modal?

kontrollanten:
Yes, it have. My bad. Shall we use `section` instead?

cromefire:
Hmm I didn't know that dialog had that ability, I thought it's just some made up tag

prateekbh:
Why not div with proper aria?

I know mdc web takes care of accessibility so I don't think it's s problem

kontrollanten:
According to https://github.com/dequelabs/axe-core/blame/develop/lib/commons/aria/index.js#L279 it's not allowed, which gives me problem in my [nightwatch-accesibility](nightwatch-accessibility) tests (built upon [axe-core](https://www.npmjs.com/package/axe-core)).

prateekbh:
Hmm then may section of the second best option.

PR #1295, opened 2019-03-05
Introducing Applitools and Cypress

### LINKED ISSUE #1282: Feature Suggestion: Applitools Visual Testing
Hey Folks, my name is James Spivey. I work with This Dot and we have teamed up with Applitools to offer free visual testing to great open source projects such as this one. Even better, they want to support helping women juniors through an apprenticeship program to do this work. This Dot would handle the pull request for this issue through the Apprenticeship program.

If you're not familiar with visual regression testing there's a pretty good summary [in the Storybook docs](https://github.com/Semantic-Org/Semantic-UI-React/issues/1604) (most of the page is unrelated to Storybook) but the tl;dr it involves rendering components individually in the DOM and taking a snapshot of the pixels and doing comparisons of it against baselines.

Most solutions do normal 1:1 pixel comparing, but this often results in false positives because the exact pixels can vary. e.g. different GPUs render things differently, so your local tests will fail against baselines created on another computer, or if the CI changes its GPU or there's a browser update, etc, etc, etc. Applitools on the other hand tries to visually compare like a human would, ignoring differences that a human wouldn't perceive or care about. The whole AI thing and such.

Maintaining a separate set of visual tests can could be a pain so what I am thinking instead was take advantage of the wonderful existing docs setup this project has. Since this project already has the ability to render the component examples all by themselves e.g.https://material.preactjs.com/component/button/ this could make rendering and testing the components easy.

My suggested solution would be to integrate Cypress and Applitools to load the docs, render the examples, and snapshot test the dom elements created from the docs.

Looking forward to talking further about this idea and the value it could add to Preact Material!

megsachdev:
- Fixes #1282

Hello!
 At times maintaining separate visual test cases becomes a tiring task so using your existing doc setup we tried coming up with something that can ease this process. So I chose to use [Cypress](https://www.cypress.io/) as the browser test harness, which Applitools has an SDK for. I started off with writing tests for some of the components so it's sort of a "smoke test" but you can add more if you'd like. 

 Visual regression involves comparing the pixels visually like a human would, ideally ignoring differences that a human couldn't perceive e.g. 1 pixel changes, differences in how browsers render images, drop shadows, etc. to avoid false positives; but you can still do a strict 1:1 pixel match if that's your preference.

Applitools thankfully  [provides an SDK](https://applitools.com/tutorials/cypress.html)  that made it trivial to write tests using Cypress. So other than the things I included what you'll need to do is  [create your own Applitools account](https://applitools.com/users/register)  and  [then setup your GitHub CI integration](https://applitools.com/docs/topics/integrations/github-integration.html)  if you want it to run automatically on Pull Requests; optional, but probably a good idea.

To run Applitools:

1.  Create Applitools account
2.  Go to the menu in the top right corner and click the profile icon on the far right.  
    [screen shot 2019-02-26 at 11 53 09 am](https://user-images.githubusercontent.com/30393329/53442046-30181e80-39bd-11e9-959b-1d08175435c8.png)
3.  Click on "My API Key" and copy your key
4.  Go to your terminal and paste  
    `export APPLITOOLS_API_KEY=Your_API_Key_Here`
5.  Go to your directory and run:  
    `yarn test:travis`
6.  In your Applitools dashboard you should see the results of your tests

sample for homepage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,7 @@ docs/build
 
 # lerna
 *.lerna_backup
+
+#cypress
+packages/docs/cypress/videos
+packages/docs/cypress/screenshots
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,13 +37,3 @@ deploy:
 branches:
   except:
     - gh-pages
-
-addons:
-  #firefox: "latest"
-  chrome: "stable"
-
-notifications:
-  webhooks:
-    urls:
-      - "https://artipost.io/travis/hook"
-    on_start: always
--- a/packages/docs/applitools.config.js
+++ b/packages/docs/applitools.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  concurrency: 50
+};
--- a/packages/docs/cypress.json
+++ b/packages/docs/cypress.json
@@ -1,0 +1,4 @@
+{
+  "chromeWebSecurity": false,
+  "eyesTimeout": 180000
+}
--- a/packages/docs/cypress/integration/visual-test/button-test.spec.js
+++ b/packages/docs/cypress/integration/visual-test/button-test.spec.js
@@ -1,0 +1,23 @@
+import {testScreenshot, openEyes, baseURL} from '../../utils/utils';
+describe('Visual  Regression Tests', () => {
+  const url = `${baseURL}/component/button`;
+  beforeEach(() => {
+    cy.visit(url);
+  });
+  afterEach(() => {
+    cy.eyesClose();
+  });
+  testScreenshot('buttonpage');
+
+  it('check all buttons', () => {
+    var buttons = cy.get('.mdc-button:enabled');
+    buttons.each(button => {
+      cy.wrap(button).click();
+    });
+    openEyes('buttons-clicked');
+    cy.eyesCheckWindow({
+      sizeMode: 'selector', //mode
+      selector: '.content.mdc-layout-grid'
+    });
+  });
+});
--- a/packages/docs/cypress/integration/visual-test/card-test.spec.js
+++ b/packages/docs/cypress/integration/visual-test/card-test.spec.js
@@ -1,0 +1,11 @@
+import {testScreenshot, openEyes, baseURL} from '../../utils/utils';
+describe('Visual  Regression Tests', () => {
+  const url = `${baseURL}/component/card`;
+  beforeEach(() => {
+    cy.visit(url);
+  });
+  afterEach(() => {
+    cy.eyesClose();
+  });
+  testScreenshot('cardpage');
+});
--- a/packages/docs/cypress/integration/visual-test/dialog-test.spec.js
+++ b/packages/docs/cypress/integration/visual-test/dialog-test.spec.js
@@ -1,0 +1,30 @@
+import {testScreenshot, openEyes, baseURL} from '../../utils/utils';
+describe('Visual  Regression Tests', () => {
+  const url = `${baseURL}/component/dialog`;
+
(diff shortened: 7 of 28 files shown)


prateekbh:
Seems like the secret env variables are [not exposed](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings) to PRs coming from forks.
In this case how will the test run for the PRs like this(coming from forks)

cc: @megsachdev @myspivey @ladyleet

myspivey:
> Seems like the secret env variables are [not exposed](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings) to PRs coming from forks.
> In this case how will the test run for the PRs like this(coming from forks)
> 
> cc: @megsachdev @myspivey @ladyleet

So this looks to be an interesting choice that TravisCI made (I have mainly dealt with CircleCI which does not have this issue) where you can not use encrypted variables for external forks. You can see that detail here, https://docs.travis-ci.com/user/pull-requests/. The option here is to disable the visual tests for external forks which is clearly not as ideal. I was able to find a resource where Algolia got this working, https://blog.algolia.com/travis-encrypted-variables-external-contributions/ but it involves some more setup.

prateekbh:
@myspivey the reason algolia server works is that they generate a very short lived new key for every request. If this kind of API, for key generation is available from APPLITOOLS then we're good to go 👍 
else its really a problem.

If APPLITOOLS can take this as a feature request then I can just leave the key unencrypted for now(as there is no billing involved anyways 😛). If not I am not sure how APPLITOOLS and open source repos can work as Travis is a major player in open source space.

cromefire:
> So this looks to be an interesting choice that TravisCI made

The reason is because a pr can be opened by every one and so there can be malicious code that reads out the secret

I have to look it up, but I think there was a way that Travis can proof to 3dr parties that it is a build of a certain repository.

prateekbh:
@cromefire to rescue 🦸

cromefire:
> https://blog.algolia.com/travis-encrypted-variables-external-contributions/

That was also the best idea I could find to prove your identity to someone from inside travis. But this would involve APPLITOOLS. Do they have some kind of issue tracker?

prateekbh:
I can easily write a super https://code.stdlib.com/ API to do the algolia server only if APPLITOOLS can provide us an API to generate multiple short lived keys.

@myspivey @ladyleet can you as them if this is possible?

myspivey:
@prateekbh on it and will get back to you with our findings as soon as possible. Sorry for the frustration here!

prateekbh:
no worries :) Thanks for the co-operation.

cromefire:
That's certainly a problem that many of this solutions have

prateekbh:
trying this out.. by not so great methods, I had a conversation with dev rel of applitools they are tring to figure it out

prateekbh:
Merging this will try to run these in parallel. @myspivey @megsachdev any ideas for running these in parallel?

if not we can run selective tests

myspivey:
> Merging this will try to run these in parallel. @myspivey @megsachdev any ideas for running these in parallel?
> 
> if not we can run selective tests

You should be able to run these with a higher concurrency. Right now it looks set to 5, although its possible for some reason it is not seeing the config file. You can set `APPLITOOLS_CONCURRENCY=100 ` as an env variable which should take precedence over that file. Give that a shot.

cromefire:
@myspivey Do you know how I may access the results of the regression tests?

prateekbh:
@cromefire DM me ur emailid, i need to add you as team member

Issue #1293, opened 2019-03-04
Standard API for major event handlers in 2.0


prateekbh:
While we're writing components for 2.0, we should standardize the API got event handlers.

Currently with props all event handlers are native event handlers.
E.g: `onChange` of a checkbox is the native onChange of `input type=checkbox`
This leads to some problems, mdc-web introduce some additional properties on top of native controls.
E.g. checkbox has a `indeterminate` state, select has `selectedIndex` etc.

Getting these values from native event handlers which just pass the `Event`, is tricky and most times requires creating a ref.

In order to solve this we can send added details to major event handlers(not necessarily all) like onChange for checkbox, select, onClick for button etc.

Current TS definition:
```
onClick: (e:Event) => void;
```

Updated version:

```
onChange:(e:Event, object: {
  MDComponent: MaterialComponent;
  checked: boolean;
  indeterminate: boolean;
}) => void
```

The additional object makes sure that it is not counter intuitive to users if 1.0

cromefire:
There currently carnival were I am, so I'm not available today

cromefire:
But that's a good idea

prateekbh:
No worries... It isn't something to be fixed immediately. That's just something we confirm to as we update all the components.. 
Enjoy the carnival ☺️

cromefire:
I guess this is done is'nt it?

prateekbh:
Tabs are remaining. 1 last one i guess

cromefire:
Possible

PR #1302, opened 2019-03-28
Dialog2.0

### LINKED ISSUE #1249: 2.0 - Update Dialog
This is a tracking issue for updating Dialog

**Scope:** Dialog

prateekbh:
Fixes #1249
Confirms to #1293 
- Adds `onClose` prop.
- Adds `onOpen` prop.
- Fixes Exports.
- Removes dependency on `@bind` because with arrow functions `module` build needs no transpilation at all. Where as with `@bind` it adds some code which will be in module build as well


packages/dialog/src/action.tsx
```diff
@@ -0,0 +1,34 @@
...
+  protected componentName = 'dialog__button';
+  protected mdcProps = [];
+
+  protected materialDom(allProps) {
+    const {accept, cancel, children, props} = allProps;
+    let actionString = 'false';
+    if (accept) {
+      actionString = 'accept';
+    } else if (cancel) {
+      actionString = 'close';
```
cromefire: Just to be sure, this are the only options right?

prateekbh: They can be anything but these are the default ones

cromefire: Shouldn't the user have a choice then?

prateekbh: They have a choice by adding a prop, but I guess we should create a`action` prep to make it easier?

cromefire: Yes and then just replace `false` with
```typescript
action ? action : 'false' 
```

cromefire: (and add it to the docs)

PR #1307, opened 2019-04-17
upgrading list

### LINKED ISSUE #1259: 2.0 - Update List
This is a tracking issue for updating List

**Scope:** List

### LINKED ISSUE #1293: Standard API for major event handlers in 2.0

prateekbh:
- Fixes #1259 
- confirms to #1293

### DIFF
--- a/packages/list/package.json
+++ b/packages/list/package.json
@@ -31,13 +31,13 @@
     "url": "https://github.com/prateekbh/preact-material-components/issues"
   },
   "dependencies": {
-    "@material/list": "0.39.3",
+    "@material/list": "1.1.1",
     "@preact-material-components/base": "^2.0.1-alpha.0",
     "@preact-material-components/icon": "^2.0.1-alpha.0",
     "bind-decorator": "^1.0.11"
   },
   "peerDependencies": {
-    "preact": "^8.3.1"
+    "preact": "^8.4.2"
   },
   "sideEffects": [
     "*.scss"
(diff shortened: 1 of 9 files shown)


packages/list/package.json
```diff
@@ -31,13 +31,13 @@
     "url": "https://github.com/prateekbh/preact-material-components/issues"
   },
   "dependencies": {
-    "@material/list": "0.39.3",
+    "@material/list": "1.1.1",
```
cromefire: Do all components have the same version? I saw some with `1.1.0`

cromefire:
I hope I have some time in the next few day, so I can help some more

packages/list/package.json
prateekbh: There might have been a release recently we'll upgrade all in the end once

cromefire: That's fine

PR #1314, opened 2019-05-16
Switch2.0

### LINKED ISSUE #1266: 2.0 - Update Switch
This is a tracking issue for updating Switch

**Scope:** Switch

prateekbh:
fixes #1266

### DIFF
--- a/packages/switch/src/index.tsx
+++ b/packages/switch/src/index.tsx
@@ -17,15 +17,6 @@ export class Switch extends MaterialComponent<ISwitchProps, ISwitchState> {
     super.componentDidMount();
     if (this.control) {
       this.MDComponent = new MDCSwitch(this.control);
-      if (this.props.value) {
-        this.MDComponent.value = this.props.value;
-      }
-    }
-  }
-
-  public componentWillUpdate(nextProps: ISwitchProps) {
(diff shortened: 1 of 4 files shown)


packages/switch/src/index.tsx
```diff
@@ -17,15 +17,6 @@ export class Switch extends MaterialComponent<ISwitchProps, ISwitchState> {
...
       this.MDComponent = new MDCSwitch(this.control);
-      if (this.props.value) {
-        this.MDComponent.value = this.props.value;
-      }
-    }
-  }
-
-  public componentWillUpdate(nextProps: ISwitchProps) {
-    if (nextProps.value && this.props.value !== nextProps.value) {
-      this.MDComponent.value = nextProps.value;
```
cromefire: Why was this removed?

prateekbh: `value` setter is removed: https://github.com/material-components/material-components-web/tree/master/packages/mdc-switch#mdcswitch-properties-and-methods

cromefire: We should maybe think of an alternative

prateekbh: shouldn't `checked` be fine?

cromefire: yes it should

PR #1319, opened 2019-06-06
Textfield2.0

### LINKED ISSUE #1269: 2.0 - Update Textfield
This is a tracking issue for updating Textfield

**Scope:** Textfield

prateekbh:
fixes #1269 

- should have more tests

### DIFF
--- a/packages/textfield/src/input.tsx
+++ b/packages/textfield/src/input.tsx
@@ -74,7 +74,11 @@ export class TextFieldInput extends MaterialComponent<
   }
 
   public componentWillUpdate(nextProps: ITextFieldInputProps) {
-    if (nextProps.value && this.props.value !== nextProps.value) {
+    if (
+      nextProps.value &&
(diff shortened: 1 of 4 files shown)


packages/textfield/src/input.tsx
```diff
@@ -131,7 +135,7 @@ export class TextFieldInput extends MaterialComponent<
             {...props}
           />
         )}
-        {props.label && this.state.jsComponentAttached && (
+        {props.label && this.state.jsComponentAttached && !outlined && (
```
cromefire: Shouln't the label be also there, when it is outlined?

prateekbh: That's: https://github.com/prateekbh/preact-material-components/pull/1319/files#diff-8e917426ccc6e2794508642f1db54718R149

cromefire: Ah allright

PR #1323, opened 2019-06-20
Linear layout 2.0

### LINKED ISSUE #1257: 2.0 - Update LayoutGrid
This is a tracking issue for updating LayoutGrid

**Scope:** LayoutGrid
**Docs:** https://material.io/develop/web/components/layout-grid/

cromefire:
Fixes #1257

Linear layout is not yet available in 1.X or 2.X

### DIFF
--- a/packages/layout-grid/package.json
+++ b/packages/layout-grid/package.json
@@ -17,6 +17,8 @@
   "license": "MIT",
   "sass": "sass",
   "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "typings": "lib/index.d.ts",
   "directories": {
     "lib": "lib",
     "test": "__tests__"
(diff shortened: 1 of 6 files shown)


cromefire:
@prateekbh where is that "login" page in applitools coming from?

packages/layout-grid/package.json
```diff
@@ -17,6 +17,8 @@
   "license": "MIT",
   "sass": "sass",
   "main": "lib/index.js",
+  "types": "lib/index.d.ts",
```
prateekbh: ohhh we neeed to do this for every other package as well

cromefire: Yes will do Search+Replace

cromefire:
Note that I want to leave this open until the base packages has been updated

prateekbh:
@cromefire re: Login screen: I was just messing around some stuff...

prateekbh:
oh I thought you wanna do that as a separate PR?

packages/layout-grid/package.json
cromefire: Part of #1325

prateekbh:
@cromefire I dont think the upgrade to base packages is coming anytime soon. should we just ship this with 0.41?

cromefire:
If you think that's a good idea. I didn't see any deprecation notice so I don't know, if there is something happening.

prateekbh:
I guess it should be fine

PR #1325, opened 2019-06-20
Image list 2.0 & doc improvements

### LINKED ISSUE #1256: 2.0 - Update ImageList
This is a tracking issue for updating ImageList

**Scope:** ImageList

cromefire:
Fixes #1256

(plus some documentation improvements)

### DIFF
--- a/packages/docs/src/components/app/hljs-darcula.scss
+++ b/packages/docs/src/components/app/hljs-darcula.scss
@@ -1,0 +1,78 @@
+/*
+
+Darcula color scheme from the JetBrains family of IDEs
+
+Copied from hljs
+
+*/
+
+.hljs {
+  display: block;
+  overflow-x: auto;
+  padding: 0.5em;
+  background: #2b2b2b;
+}
+
+.hljs {
+  color: #bababa;
+}
+
+.hljs-strong,
+.hljs-emphasis {
+  color: #a8a8a2;
+}
+
+.hljs-bullet,
+.hljs-quote,
+.hljs-link,
+.hljs-number,
+.hljs-regexp,
+.hljs-literal {
+  color: #6896ba;
+}
+
+.hljs-code,
+.hljs-selector-class {
+  color: #a6e22e;
+}
+
+.hljs-emphasis {
+  font-style: italic;
+}
+
+.hljs-keyword,
+.hljs-selector-tag,
+.hljs-section,
+.hljs-attribute,
+.hljs-name,
+.hljs-variable {
+  color: #cb7832;
+}
+
+.hljs-params {
+  color: #b9b9b9;
+}
+
+.hljs-string {
+  color: #6a8759;
+}
+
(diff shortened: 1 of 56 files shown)


packages/docs/src/components/app/hljs-darcula.scss
```diff
@@ -0,0 +1,79 @@
+/*
```
prateekbh: why is this needed?

cromefire:
> ^ shows that code was highlighted better earlier but now its all black. Expected?

Sorry a bit malfunctional css

packages/docs/src/components/app/hljs-darcula.scss
cromefire: For the dark code theme, white code in the dark theme looks ugly

cromefire:
@prateekbh This is now ready, dark docs improvements (e.g. the site now saves your settings), test improvements, image list upgrade and a few smaller things

packages/docs/src/components/app/index.js
```diff
@@ -196,9 +199,19 @@ const menuItems = [
...
   constructor() {
     super();
+
+    // Dark mode
+    const darkMode =
+      typeof localStorage !== 'undefined'
+        ? localStorage.getItem('dark-mode') === 'dark'
+        : false;
+    if (darkMode) {
+      document.body.classList.add('mdc-theme--dark');
```
prateekbh: NIT: may be we can just click on the switch instead?

cromefire: Yes the switch is still there, but each time you reload the page you have to put it into dark mode again. Now it's the same as before, but it remembers the users choice

cromefire:
> Just a request I can really use smaller PR reviews though neutral_face

Yes I try but when you fix one thing, you get the next bug/issue and so on and shortly you have a bunch of fixes

This time I just improved the image list page and then fixed all the bugs caused by it

packages/docs/src/components/app/index.js
cromefire: I personally find it really annoying when page becomes white again suddenly and as it an easy fix I implemented it. What do you think?

cromefire:
Right now there only a few components left and then some structural thing that we have to decide if we want to tackle them in `2.0`

cromefire:
And I guess all the greenkeeper branches could be delete, because there are like 50+ useless branches

packages/docs/src/components/app/index.js
prateekbh: oh yeah what i mean is should we just `$(switch).click()` instead of doing this with classes?

cromefire: The only thing we would save is this line:
```typescript
document.body.classList.add('mdc-theme--dark');
```

And that approach feels kind of _cleaner_ to me, I don't like UI elements that do things on their own

Issue #1250, opened 2019-01-26
2.0 - Update Drawer


prateekbh:
This is a tracking issue for updating Drawer

**Scope:** Drawer

cromefire:
Merged into 2.0

PR #1326, opened 2019-07-01
Upgrading drawer

### LINKED ISSUE #1251: 2.0 - Add gesture support to drawer
We need to add support for swipe gesture to our modal drawer, because its 2019.

prateekbh:
Fixes #1251 #1250 

- Adds gesture support for drawer
- Upgrades Drawer
- Confirms to #1293 

PMCSlider


cromefire:
Nice one

packages/drawer/src/index.tsx
```diff
@@ -15,11 +16,32 @@ export interface IDrawerProps extends JSX.HTMLAttributes {
...
+export interface IDrawerState {
+  sliderState: SLIDER_STATES;
+  touchPositionX: number;
+  touchPositionY: number;
+  swipeDistance: number;
+}
+
+const CLOSED_STATE = {
+  sliderState: SLIDER_STATES.REST,
+  swipeDistance: -256,
```
cromefire: Is this okay to be hardcoded?

packages/drawer/src/index.tsx
```diff
@@ -55,13 +82,105 @@ export class Drawer extends MaterialComponent<IDrawerProps, IDrawerState> {
...
+          {...props}
+          style={styleObject}>
           {props.children}
         </aside>
         {props.modal && <div class="mdc-drawer-scrim" />}
       </div>
     );
   }
+
+  private handleTouchStart_ = e => {
```
cromefire: Why do they all have an underscore at the end?

packages/drawer/src/index.tsx
```diff
@@ -55,13 +82,105 @@ export class Drawer extends MaterialComponent<IDrawerProps, IDrawerState> {
...
+    const startPosition = this.MDComponent.open ? 0 : -256;
+    const {clientX, clientY} = touches[0];
+    const deltaX = clientX - this.state.touchPositionX;
+    this.setState({
+      sliderState: SLIDER_STATES.SWIPING,
+      swipeDistance: startPosition + deltaX
+    });
+  };
+
+  private attachGlobalListeners_ = () => {
```
cromefire: Are they cleaned up again?

cromefire:
Is there any way that we can automatically test this? I guess we'll have to look into drag & drop support

prateekbh:
@cromefire yeah we just need to enable this for mousedown/up/move and we'll be able to test it with cypress

prateekbh:
@cromefire Tests for swipe added

prateekbh:
ping

cromefire:
There are still 3 comments left unanswered

packages/drawer/src/index.tsx
prateekbh: we dont really have a good way of detecting this... so our best bet are our visual tests

packages/drawer/src/index.tsx
prateekbh: cuz they are private and we ship js not TS. WDYT?

packages/drawer/src/index.tsx
prateekbh: missed that

packages/drawer/src/index.tsx
cromefire: Most IDEs use the .d.ts file also to suggest things in JavaScript (at least WebStorm and VSCode I think)

packages/drawer/src/index.tsx
cromefire: We will see

packages/drawer/src/index.tsx
prateekbh: i know but 
a. lets not take that for granted
b. does it hurt? even MDC-web ships this way

cromefire: It's not consistent with the rest of the codebase and I think it's quite ugly, but if you really want to do this

cromefire:
Also since when do we have netlify?

cromefire:
I think after this there is not that much to do until we can roll the first alpha 🎉

cromefire:
This does not look correct:

Screenshot

prateekbh:
yep thats y added a preview deploy from netlify also the navigation links are not really working

packages/drawer/src/index.tsx
prateekbh: yeah i really dont have a good way of calculating this right now

packages/drawer/src/index.tsx
prateekbh: i'll remove it

prateekbh:
@cromefire all known corner cases are now covered.

I'll work on some cross browser issues post alpha release

cromefire:
That looks quite goo but let us look at 2
Short things:
- The drawer is only visible if I Touch it
- I think we should extract that into it's own class instead using a prop (can do that)

prateekbh:
Gotcha... Will do in next release

PR #1335, opened 2019-09-17
PreactX support

prateekbh:
- Support for preact X in 2.0

### DIFF
--- a/packages/base/src/MaterialComponent.ts
+++ b/packages/base/src/MaterialComponent.ts
@@ -85,7 +85,6 @@ export abstract class MaterialComponent<
     } else {
       element.attributes = element.attributes || {};
     }
-
     // @ts-ignore
     element[propName].className = `${userDefinedClasses} ${this.getClassName(
       element
--- a/packages/base/src/MaterialComponent.ts
+++ b/packages/base/src/MaterialComponent.ts
@@ -76,9 +76,18 @@ export abstract class MaterialComponent<
     }
 
     const element = this.materialDom(componentProps);
-    element.attributes = element.attributes || {};
+    let propName = 'attributes';
 
-    element.attributes.className = `${userDefinedClasses} ${this.getClassName(
+    if ('props' in element) {
+      propName = 'props';
+      // @ts-ignore
+      element.props = element.props || {};
+    } else {
+      element.attributes = element.attributes || {};
+    }
+
+    // @ts-ignore
+    element[propName].className = `${userDefinedClasses} ${this.getClassName(
       element
     )}`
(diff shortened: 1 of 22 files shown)


prateekbh:
@cromefire  when ever i try to build docs I am getting the following error 

https://pastebin.com/4aWwCV7n

I remember you fixed this earlier, any ideas?

cromefire:
Yes I wrote a custom resolver for `sass-loader`:

https://github.com/prateekbh/preact-material-components/blob/762edebc0ab1f340b693bed6cba98ff29daa9297/packages/docs/preact.config.js#L11-L25

https://github.com/prateekbh/preact-material-components/blob/762edebc0ab1f340b693bed6cba98ff29daa9297/packages/base/build/index.ts#L1-L124

packages/docs/preact.config.js
```diff
@@ -1,4 +1,4 @@
-const {sassResolver} = require('@preact-material-components/base/build');
+const {sassResolver} = require('@preact-material-components/base');
```
cromefire: This might be the problem

cromefire (changes requested):

packages/docs/preact.config.js
prateekbh: oh yeah that makes sense

cromefire (approved):

prateekbh:
https://github.com/prateekbh/preact-material-components/issues/1336 - noted failing tests here

cromefire:
Why did you do the `dummy commit`?

prateekbh:
should've just restarted the tests 🤦🏻‍♂️

packages/base/src/MaterialComponent.ts
```diff
@@ -76,23 +76,33 @@ export abstract class MaterialComponent<
     }
 
     const element = this.materialDom(componentProps);
-    element.attributes = element.attributes || {};
-
-    element.attributes.className = `${userDefinedClasses} ${this.getClassName(
+    let propName = 'attributes';
```
developit: Suggestion: throw this into a utility to avoid repeated lookups and dynamic property access:

```js
import { h } from 'preact';

const isX = 'props' in h('a');

export const getProps = isX ? vnode => vnode.props : vnode => vnode.attributes;

export const setProps = isX ? (vnode, props) => vnode.props = props : (vnode, props) => vnode.attributes = props;
```

developit: (see https://gist.github.com/developit/17fdda2e2236d0c390e8a9bd947f044d)

PR #1349, opened 2020-01-13
fix: sizeMode => target

TroyTae:
In cypress, sizeMode is deprecated.
So I replaced all of them to target.


prateekbh:
Amazing work @TroyTae Thanks so much

